PR #6, opened 2025-07-07
fix wrong crawl format ('html' to 'rawHtml')

XHyperDEVX:
it fixes the bug that prevents some users from using local firecrawl-simple instances. as found out [here](https://discord.com/channels/1086345563026489514/1389590707949273148/1391439767597613229), it is because librechat sends “html” as format, which was removed by firecrawl-simple in december 2024.

### DIFF
--- a/src/tools/search/firecrawl.ts
+++ b/src/tools/search/firecrawl.ts
@@ -23,7 +23,7 @@ export class FirecrawlScraper {
       'https://api.firecrawl.dev';
     this.apiUrl = `${baseUrl.replace(/\/+$/, '')}/v1/scrape`;
 
-    this.defaultFormats = config.formats ?? ['markdown', 'html'];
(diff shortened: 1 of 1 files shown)


michnovka:
I would prefer to make this configurable. And ideally, we should also allow specifying other stuff, such as `headers`. My primary use case for this would be to specify custom user-agent.

But there are so many - https://docs.firecrawl.dev/api-reference/endpoint/crawl-post#body-scrape-options

Perhaps we can pass a JSON array if you don't want to make a config option for each?

but:

- removeBase64Images
- blockAds
- location
- parsePdf

...

all these look really useful!

XHyperDEVX:
I agree, more setting options for this would be nice. However, I see this more as an “add-on”. In my opinion, the feature itself should work first. Only when this is the case can you continue to work on it and add further options.

danny-avila:
Closed in favor of #9

PR #11, opened 2025-08-11
feat: allow configurable Cohere API URL in CohereReranker

davidjrh:
This pull request makes a small but important change to the `CohereReranker` class to improve configurability of the Cohere API endpoint.

- The URL used for the Cohere rerank API in `CohereReranker` can now be overridden by setting the `COHERE_URL` environment variable; if not set, it defaults to the original Cohere endpoint.

This allows to specify a different base URL for Cohere, such as the ones provided by the Azure AI Foundry.


danny-avila:
https://github.com/danny-avila/agents/pull/498

PR #23, opened 2025-09-28
fix: No stepId for tool_call_id issues

aldehir:
I've been trying to find the source of the following bug that occurs in LibreChat when using llama.cpp and `gpt-oss-20b`:

image

I believe I have identified the reason why this happens.

With llama.cpp, the first tool call chunk contains partial arguments for a tool call:

```
data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"id":"VPwsTAer3qtRES726PlBwABM7aRJAyIC","type":"function","function":{"name":"calculator","arguments":"{\""}}]}}],"created":1759034187,"id":"chatcmpl-ZJ1BpOZ5nguZXOCkFsqLybn2N6A76kLx","model":"gpt-oss-20b","system_fingerprint":"b6590-f00a36081","object":"chat.completion.chunk"}
```

However, with LM Studio, the first tool chuck contains no arguments.

```
data: {"id":"chatcmpl-p9z8bcrt8ps5mlflzmbii","object":"chat.completion.chunk","created":1759033266,"model":"gpt-oss-20b@f16","system_fingerprint":"gpt-oss-20b@f16","choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"id":"682997534","type":"function","function":{"name":"calculator","arguments":""}}]},"logprobs":null,"finish_reason":null}]}
```

The partial arguments are an issue because langchain does not populate `chunk.tool_calls` if it cannot parse the arguments. Instead, it will place the tool call in `chunk.invalid_tool_calls` as shown here:

image

An empty set of arguments, however, parse to an empty object `{}`. As show here when using LM Studio:

image

Notice how it populates `chunk.tool_calls` in this instance.

By adding a check for tool call chunks in `handleReasoning()`, I can get it to work:

image

Unfortunately, the problems don't quite stop there. The next issue is if a streaming message contains both content and a tool call, which several models can do:

image

I believe the issue lies somewhere in the `handleToolCallChunks()` logic:

https://github.com/danny-avila/agents/blob/d497aa17581e0edd437fe86c23c87c92c21452d9/src/tools/handlers.ts#L36

`graph.contentData.length` is usually larger than `graph.stepKeyIds[stepKey].length` after a couple of steps. I'll be honest, I don't quite understand the overall logic here, but the end result is `prevStepId == undefined`, which stops it from adding the tool call. By removing the second argument, so it returns the last step id for the given step key, it appears to work well:

image

@danny-avila, I'd love to know what you think about this issue.

### DIFF
--- a/src/tools/handlers.ts
+++ b/src/tools/handlers.ts
@@ -133,7 +133,7 @@ export const handleToolCalls = async (
     let prevStepId = '';
     let prevRunStep: t.RunStep | undefined;
     try {
-      prevStepId = graph.getStepIdByKey(stepKey, graph.contentData.length - 1);
+      prevStepId = graph.getStepIdByKey(stepKey);
       prevRunStep = graph.getRunStep(prevStepId);
     } catch {
       // no previous step


aldehir:
I have been using this for quite some time, both with llama.cpp and an Azure OpenAI deployment. Nothing seems broken and I haven't seen the stepId issue since, so I'm removing it from the draft state.

Searching previous discussions, these comments appear relevant: https://github.com/danny-avila/LibreChat/discussions/8545#discussioncomment-14150730, although the original discussion is likely not related.

PR #29, opened 2025-11-10
feat: added Tavily integration as both a search provider and a scraper prov…

MayRamati:


danny-avila:
Closed in favor of https://github.com/danny-avila/agents/pull/85

PR #47, opened 2026-01-13
feat: add COHERE_API_URL support for custom Cohere rerank endpoints

busla:
## Summary
- Allow configuring a custom Cohere API URL via the `COHERE_API_URL` environment variable
- Falls back to the default `api.cohere.com/v2/rerank` endpoint when not set
- Mirrors the existing `JINA_API_URL` pattern for JinaReranker

## Motivation
Users want to use Cohere-compatible reranking endpoints like LiteLLM proxy or other self-hosted solutions. Currently, the Cohere reranker URL is hardcoded, preventing this use case.

Related: https://github.com/danny-avila/LibreChat/pull/5944

## Changes
- Add `apiUrl` property to `CohereReranker` class
- Add `apiUrl` parameter to `CohereReranker` constructor with env var fallback
- Update `rerank` method to use dynamic `apiUrl` instead of hardcoded URL
- Add `cohereApiUrl` parameter to `createReranker` factory function

## Test plan
- Set `COHERE_API_URL` to a custom endpoint and verify requests go there
- Leave `COHERE_API_URL` unset and verify fallback to default URL works

Dependency: https://github.com/danny-avila/LibreChat/pull/9544

### DIFF
--- a/src/tools/search/cohere-reranker.test.ts
+++ b/src/tools/search/cohere-reranker.test.ts
@@ -1,0 +1,127 @@
+import { CohereReranker, createReranker } from './rerankers';
+import { createDefaultLogger } from './utils';
+
+// Helper to access private apiUrl property for testing
+const getApiUrl = (reranker: CohereReranker): string =>
+  (reranker as unknown as { apiUrl: string }).apiUrl;
+
+describe('CohereReranker', () => {
+  const mockLogger = createDefaultLogger();
+
+  describe('constructor', () => {
+    it('should use default API URL when no apiUrl is provided', () => {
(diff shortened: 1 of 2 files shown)


busla:
Superseding this with #498 — same change, rebased and with test coverage mirroring the JinaReranker suite. Companion to danny-avila/LibreChat#15518 on the LibreChat side.

Also worth noting for anyone watching: #11 (@davidjrh, Aug 2025) and #149 (@npeham, May 2026) propose the same `apiUrl`/`COHERE_API_URL` option independently, both still open and unanswered. Closing this one to consolidate effort, not because the need went away.

PR #52, opened 2026-02-08
feat: add staticInstructions for multi-block Anthropic prompt cache o…

realjimhamilton:
…ptimization

Adds optional staticInstructions field to AgentInputs that enables separating static system instructions from dynamic content for Anthropic prompt caching.

When staticInstructions is provided and promptCache is true, buildSystemRunnable() creates two content blocks: the static portion with cache_control (cached and shared across users) and the dynamic-only remainder without cache_control (sent fresh each request).

This dramatically improves cache hit rates in multi-user deployments where agents have large static base prompts (~90K tokens) but small per-user dynamic context (~5-15K tokens).

Changes:

- Add staticInstructions to AgentInputs interface (types/graph.ts)

- Add staticInstructions property, constructor param, and fromConfig() (AgentContext.ts)

- Add multi-block cache logic in buildSystemRunnable() with dynamic content extraction

- Fully backwards compatible: field is optional, single-block behavior unchanged when not provided

### DIFF
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -37,6 +37,7 @@ export class AgentContext {
       toolRegistry,
       toolDefinitions,
       instructions,
+      staticInstructions,
       additional_instructions,
       streamBuffer,
       maxContextTokens,
@@ -57,6 +58,7 @@ export class AgentContext {
       toolRegistry,
       toolDefinitions,
       instructions,
+      staticInstructions,
       additionalInstructions: additional_instructions,
       reasoningKey,
       toolEnd,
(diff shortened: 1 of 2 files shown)


danny-avila:
Thanks for the contribution here. We ended up implementing this path more broadly in #128, so I am going to close this PR as superseded.

This PR introduced a new `staticInstructions` field and split Anthropic system content by subtracting that text from the full instruction string. #128 keeps the existing public shape instead: `instructions` remains the stable/cacheable prefix and `additional_instructions` becomes the dynamic system tail. It also covers the cache placement details for Anthropic and Bedrock, treats cross-run summaries as dynamic, and pairs with LibreChat PR https://github.com/danny-avila/LibreChat/pull/12907 to move memory/RAG/tool/web-search/date context out of the stable prefix.

So the idea is appreciated, but #128 is the more complete implementation we will review and carry forward.

PR #55, opened 2026-02-13
fix: inject handoff instructions into last ToolMessage to avoid "user after tool" API error

### LINKED ISSUE #54: [Bug] 400 Unexpected role 'user' after role 'tool' when handoff follows a tool call
**What happens**  
When an agent runs a tool (e.g. `list_upload_sessions`) and then hands off to another agent in the same turn, the request can fail with:

`400 Unexpected role 'user' after role 'tool'`

Image

**Why**  
For the receiving agent, handoff instructions are added as a **HumanMessage** (user role). If the last message in the filtered history is a **ToolMessage** (from the tool the router just ran), the sequence becomes `… tool → user`. Many chat APIs require an assistant turn after a tool message, so they reject this order.

**When it appears**  
Only when the router runs a non-handoff tool and then does a handoff. If the router hands off without running another tool first, the last message is not a tool message and the error does not occur.

**Suggested fix**  
When building messages for the receiving agent, if the last message in `filteredMessages` is already a ToolMessage, append the handoff instructions to that message’s content instead of adding a new HumanMessage. That keeps the sequence as `… tool` (with instructions in the tool message) and the next step is the assistant (receiving agent), which is valid.

**Steps to reproduce**  
1. Start with Universal (or any router that can call tools and hand off).  
2. Ask for something that triggers a tool call and then a handoff (e.g. “I want to visualize CSV data, please provide an upload” → router runs `list_upload_sessions`, then transfers to Data Analysis).  
3. After the transfer, the 400 error can appear.

JumpLink:
Fixes #54

When the receiving agent's `filteredMessages` end with a ToolMessage (e.g. the router ran `list_upload_sessions` before the handoff), we previously appended handoff instructions as a `HumanMessage`. That produced the sequence `… tool → user`, which many chat APIs reject with `400 Unexpected role 'user' after role 'tool'`.

This change: if the last message is a `ToolMessage`, we extend its content with the handoff instructions instead of adding a new `HumanMessage`. The next turn is then a valid assistant response. If the last message is not a tool message, behaviour is unchanged (we still append a `HumanMessage` with the instructions).

Token map handling is updated so the instructions are counted in either the extended tool message or the new human message.

## Testing

Seems to be working now (compare this with the screenshot in the issue): 
grafik

### DIFF
--- a/src/graphs/MultiAgentGraph.ts
+++ b/src/graphs/MultiAgentGraph.ts
@@ -786,13 +786,51 @@ export class MultiAgentGraph extends StandardGraph {
           /** Build messages for the receiving agent */
           let messagesForAgent = filteredMessages;
 
-          /** If there are instructions, inject them as a HumanMessage to ground the agent */
+          /**
+           * Inject handoff instructions without violating API role order.
+           * Many chat APIs require: after a tool message, the next role must be assistant, not user.
+           * If we appended a HumanMessage here when the last message is already a ToolMessage (e.g.
+           * from list_upload_sessions before the handoff), we would get "400 Unexpected role 'user'
+           * after role 'tool'". So when the last message is a ToolMessage, we extend its content
+           * with the instructions instead of adding a new HumanMessage.
+           */
           const hasInstructions = instructions !== null && instructions !== '';
           if (hasInstructions) {
-            messagesForAgent = [
-              ...filteredMessages,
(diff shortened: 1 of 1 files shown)


danny-avila:
follow-up PR to retain stronger steering of the human message for task instructions: https://github.com/danny-avila/agents/pull/59/changes

PR #61, opened 2026-02-17
fix(handoff): return wrong-tool correction to LLM instead of silent execute

### LINKED ISSUE #60: Handoff tool not found: wrong tool name (duplicated suffix) when transferring to agent
### Problem

When an agent that performs handoffs (in our case a router agent we created) tries to transfer to another agent (e.g. a “Code Refactorer”), it sometimes calls the handoff tool with a **wrong name**: the correct ID suffix is duplicated (e.g. `...7yu` becomes `...7yu7yu`). The tool is then not found on the host (handoff tools are graph-managed, not in the host’s `loadTools`), which leads to repeated failed handoffs and multiple “Transferred to Agent” messages without an actual transfer.

**Observed:** Several “Transferred to Agent” lines in a row, then the assistant reports that the transfer to the code refactorer failed.

**Provider/Model:** *Scaleway / devstral-2-123b-instruct-2512*

(The full ModelSpec configuration can be viewed [here](https://github.com/faktenforum/ai-chat-interface/blob/223fe85b21bc53a552479421f23e64c46d016bd1/packages/librechat-init/config/librechat.yaml#L482) and the full Agents configuration [here](https://github.com/faktenforum/ai-chat-interface/blob/223fe85b21bc53a552479421f23e64c46d016bd1/packages/librechat-init/config/agents.yaml#L868).)

### Relevant logs

```
2026-02-17 13:41:29 warn: [ON_TOOL_EXECUTE] Tool "lc_transfer_to_agent_V_29lG4o1TnQ3-C0wu7yu7yu" not found. Available: 

2026-02-17 13:41:31 warn: [ON_TOOL_EXECUTE] Tool "lc_transfer_to_agent_V_29lG4o1TnQ3-C0wu7yu7yu" not found. Available: 

2026-02-17 13:41:32 warn: [ON_TOOL_EXECUTE] Tool "lc_transfer_to_agent_V_29lG4o1TnQ3-C0wu7yu7yu" not found. Available: 

2026-02-17 13:41:33 warn: [ON_TOOL_EXECUTE] Tool "lc_transfer_to_agent_V_29lG4o1TnQ3-C0wu7yu7yu" not found. Available: 

2026-02-17 13:41:34 warn: [ON_TOOL_EXECUTE] Tool "lc_transfer_to_agent_V_29lG4o1TnQ3-C0wu7yu7yu" not found. Available: 

2026-02-17 13:41:36 error: [api/server/controllers/agents/client.js #sendCompletion] Operation aborted 400 Cannot set `add_generation_prompt` to True when the last message is from the assistant. Consider using `continue_final_message` instead.

2026-02-17 13:41:36 error: [api/server/controllers/agents/client.js #sendCompletion] Unhandled error type 400 Cannot set `add_generation_prompt` to True when the last message is from the assistant. Consider using `continue_final_message` instead.
```

The correct tool name would be `lc_transfer_to_agent_V_29lG4o1TnQ3-C0wu7yu` (without the duplicated `7yu`).

### Screenshot

Image

### Context

- Handoff tools are created with names like `lc_transfer_to_` + `edge.to` (agent ID). Edges/init do not duplicate the ID.
- Possible mitigation: when a handoff tool name is “wrong” but prefix-matches a known handoff tool, return a ToolMessage to the LLM with the correct name so it can call the right tool in the next turn (instead of executing silently or leaving the host to report “tool not found”).

JumpLink:
When the model calls a handoff tool with a **wrong name** (e.g. duplicated suffix like `lc_transfer_to_agent_...7yu7yu` instead of `...7yu`), we no longer execute the handoff silently with the resolved name. We return a **ToolMessage** that tells the LLM the used name was wrong and suggests the correct tool name, so the LLM can call the correct tool in the next turn and select the right agent itself.

### Out of Scope?

I understand that the PR is out of scope and may simply be a problem with the model used. I have integrated it as a workaround and didn't want to leave it unmentioned and is more likely a workaround.

### Problem

- Handoff tools are graph-managed; the host’s `loadTools` does not provide them.
- If the model emits a slightly wrong tool name (e.g. duplicated suffix), the call was previously resolved via prefix match and executed as a direct handoff with the correct name.
- That hid the mistake from the model and didn’t give it a chance to correct itself.

### Solution

- **ToolNode** (`dev/agents/src/tools/ToolNode.ts`): For handoff calls that are not in `directToolNames` but can be resolved via `resolveHandoffToolName()` (prefix match), we **do not** add them to direct execution.
- Instead we create a **ToolMessage** per such call with:
  - a short explanation that the used tool name doesn’t match a registered handoff tool,
  - the name that was used,
  - the **correct** handoff tool name,
  - an instruction to call that tool instead.
- These messages are appended to the tool outputs; `handleRunToolCompletions` is called for them so the UI shows the step and the correction.
- The LLM sees the correction as a normal tool result and can issue a follow-up call with the suggested tool name.

### Changes

- **ToolNode.ts**
  - Removed execution of resolved handoffs via `directCallsResolved`.
  - Added `handoffCorrectionMessages`: one ToolMessage per resolved handoff with the text above.
  - Outputs are now `directOutputs` + `handoffCorrectionMessages` + `eventOutputs`.
  - JSDoc for `resolveHandoffToolName` updated to describe the new behaviour.
  - Explicit null/size check for `directToolNames` to satisfy lint.
- **docs/HANDOFF_TOOL_NAME_ANALYSIS.md**: Conclusion and “Defensive fix” step updated to describe “return correction message to LLM” instead of “execute as direct”.

### Related

- Fixes #60

### DIFF
--- a/src/tools/ToolNode.ts
+++ b/src/tools/ToolNode.ts
@@ -84,6 +84,29 @@ export class ToolNode<T = any> extends RunnableCallable<T, T> {
   }
 
   /**
+   * Resolves a handoff tool call name that may be slightly wrong (e.g. LLM
+   * duplicated a suffix) to the actual direct tool name. When a match is found,
+   * we return a ToolMessage to the LLM with the correct name so it can call the
+   * right tool itself, instead of executing the handoff silently or dispatching
+   * via ON_TOOL_EXECUTE (where the host has no handoff tools).
+   */
+  private resolveHandoffToolName(callName: string): string | undefined {
+    if (
+      !callName.startsWith(Constants.LC_TRANSFER_TO_) ||
+      this.directToolNames == null ||
+      this.directToolNames.size === 0
+    ) {
+      return undefined;
+    }
+    const handoffDirect = Array.from(this.directToolNames).filter((d) =>
+      d.startsWith(Constants.LC_TRANSFER_TO_)
+    );
+    const prefixMatches = handoffDirect.filter((d) => callName.startsWith(d));
+    if (prefixMatches.length === 0) return undefined;
+    return prefixMatches.reduce((a, b) => (a.length >= b.length ? a : b));
+  }
+
+  /**
    * Returns cached programmatic tools, computing once on first access.
    * Single iteration builds both toolMap and toolDefs simultaneously.
    */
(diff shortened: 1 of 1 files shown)


danny-avila:
Closing in favor of #161 as the replacement patch for #60. The edge case is real, but #161 handles it more robustly: mistyped handoff-shaped calls fail closed in ToolNode's local unknown-tool path, provide exact-name guidance, and are proven not to leak to ON_TOOL_EXECUTE or silently execute a guessed handoff. It also adds regression coverage for both plain graph handoffs and mixed event-driven host-tool flows, plus live OpenAI and claude-sonnet-4-6 handoff validation.

PR #92, opened 2026-04-13
🔍 fix: Preserve Anthropic `server_tool_use` block types

danny-avila:
## Summary
- Server tool blocks (`web_search` with `srvtoolu_` IDs) were losing their `server_tool_use` type through the message pipeline during multi-turn conversations, becoming regular `tool_use` blocks
- On the second turn, the Anthropic API rejected the request because `tool_use` requires a matching `tool_result` block, but server tools use inline `web_search_tool_result` instead
- This only reproduced when the first run had multiple web searches via the official Anthropic API (not Azure)

## Root Cause
Three places in the pipeline could degrade `server_tool_use` → `tool_use`:
1. **`modifyDeltaProperties`** (`core.ts`): The Anthropic allowed-types list was missing `server_tool_use`/`web_search_tool_result`/`web_search_result`, causing them to be converted to `text`
2. **`_formatContent`** (`message_inputs.ts`): After LangChain's `concat()` merges `server_tool_use` + `input_json_delta` by index, the type becomes `input_json_delta` → then gets normalized to `tool_use` — but was never corrected back to `server_tool_use`
3. **String content path** (`message_inputs.ts`): When `message.content` is a string, all `tool_calls` (including `srvtoolu_` ones) were converted to `tool_use` blocks with no matching result

## Changes
- **`src/messages/core.ts`**: Add `server_tool_use`, `web_search_tool_result`, `web_search_result` to Anthropic allowed types
- **`src/llm/anthropic/utils/message_inputs.ts`**: Correct `tool_use` blocks with `srvtoolu_` IDs back to `server_tool_use`; filter server tool calls from string content path
- **`src/llm/anthropic/utils/server-tool-inputs.test.ts`**: 5 tests covering all fix scenarios

## Test plan
- [x] `npx jest server-tool-inputs` — 5 new tests pass
- [x] `npx jest prune` — 48 existing tests pass  
- [x] `npx jest content.test` — 14 existing tests pass
- [x] `npx tsc --noEmit` — clean

Closes danny-avila/LibreChat#12576

### DIFF
--- a/src/llm/anthropic/types.ts
+++ b/src/llm/anthropic/types.ts
@@ -6,7 +6,7 @@ export type AnthropicMessageDeltaEvent = Anthropic.MessageDeltaEvent;
 export type AnthropicMessageStartEvent = Anthropic.MessageStartEvent;
 
 export type AnthropicToolResponse = {
-  type: 'tool_use';
+  type: 'tool_use' | 'server_tool_use';
   id: string;
   name: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/src/llm/anthropic/utils/message_inputs.ts
+++ b/src/llm/anthropic/utils/message_inputs.ts
@@ -696,9 +696,38 @@ export function _convertMessagesToAnthropicPayload(
     }
   });
   return {
-    messages: mergeMessages(formattedMessages),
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #96, opened 2026-04-14
feat: reconstruct skill body HumanMessages in formatAgentMessages

danny-avila:
Follow-up to #95 (merged).

When a skill is invoked, the body is injected as a HumanMessage into LangGraph state but NOT persisted to conversation history. On follow-up runs the skill body is lost.

\`formatAgentMessages\` now accepts an optional \`invokedSkillBodies\` Map (\`skillName → body\`) and reconstructs HumanMessages at the correct position in the message sequence — after the skill's ToolMessage, matching where ToolNode originally injected them.

Detection happens inside the existing tool_call processing loop (both with and without tools filtering), so there are **zero extra message iteration passes**.


danny-avila:
@codex review

PR #110, opened 2026-04-18
fix(summarization): log and recover from model init errors

danny-avila:
## Summary

- Move `initializeModel` inside the try/catch in `executeSummarizationWithFallback` so that errors thrown during model construction (e.g. `Unsupported LLM provider: <name>`) are surfaced through the `log('error', ...)` path and fall through to the metadata-stub fallback, instead of bubbling up silently.
- This is the defense-in-depth half of the fix for [LibreChat Discussion #12614](https://github.com/danny-avila/LibreChat/discussions/12614): when a caller forwards an unrecognized summarization provider name (e.g. a custom-endpoint label), the error is now observable in operator logs and summarization continues with a metadata stub rather than crashing the node.
- Related to [LibreChat Discussion #12733](https://github.com/danny-avila/LibreChat/discussions/12733), which reports exactly this silent-failure mode.

The companion fix in LibreChat translates custom-endpoint provider names before they reach the SDK, so most users should never see this error path — this PR is a safety net for misconfigurations and future edge cases.

## Test plan
- [x] `npx jest src/summarization/` — all 24 tests pass, including the new `catches model initialization errors and falls back to metadata stub`.
- [x] New test verifies the fix by mocking `getChatModelClass` to throw and asserting the node resolves instead of rejecting, and that `setSummary` receives the metadata stub.
- [x] `tsc --noEmit` and `eslint` clean on the modified files.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -190,8 +190,19 @@ jobs:
         if: steps.check.outputs.skip != 'true'
         run: npm pack
 
+      - name: Determine publish tag
+        if: steps.check.outputs.skip != 'true'
+        id: publish_tag
+        run: |
+          PACKAGE_VERSION=$(node -p "require('./package.json').version")
+          if echo "$PACKAGE_VERSION" | grep -q "-"; then
+            echo "tag=dev" >> $GITHUB_OUTPUT
+          else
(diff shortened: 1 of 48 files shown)


danny-avila:
@codex review

PR #111, opened 2026-04-18
🛡️ fix: validate edge agent refs early in MultiAgentGraph

danny-avila:
## Summary

Safety-net fix for the root failure mode seen in [LibreChat #12726](https://github.com/danny-avila/LibreChat/issues/12726): when a host passes `edges` into a `multi-agent` run without including every referenced agent in `agents`, `StateGraph.compile()` throws the opaque `Found edge ending at unknown node "<id>"` error — with no indication that the true cause is a missing agent config.

This change validates the edges up-front in the `MultiAgentGraph` constructor and throws a descriptive error naming every missing agent id, pointing callers at the real fix (include the agent or filter the orphan edge).

## Why in the SDK as well as the host?

LibreChat's own controllers now pre-load handoff sub-agents (see the [companion PR](https://github.com/danny-avila/LibreChat/pull/12740)), so this SDK guard is not load-bearing for the LibreChat bug. It's here because the same class of mistake is easy for any downstream consumer to make, and the LangGraph error message gives them nothing actionable. Failing fast with a clear message saves the next person a long debugging session.

## Test plan

- [x] New unit test file: `src/graphs/__tests__/MultiAgentGraph.test.ts` — 6 cases covering valid graphs, unknown `to`, unknown `from`, multiple unknowns reported together, array-form `from`/`to`, and empty-edges case.
- [x] All pass: `npx jest src/graphs/__tests__/MultiAgentGraph.test.ts`.
- [x] `npx tsc --noEmit` clean.

### DIFF
--- a/src/graphs/MultiAgentGraph.ts
+++ b/src/graphs/MultiAgentGraph.ts
@@ -59,9 +59,48 @@ export class MultiAgentGraph extends StandardGraph {
   constructor(input: t.MultiAgentGraphInput) {
     super(input);
     this.edges = input.edges;
+    this.validateEdgeAgents();
     this.categorizeEdges();
     this.analyzeGraph();
     this.createHandoffTools();
+  }
+
+  /**
+   * Fails fast when an edge references an agent that is not in
+   * `agentContexts`. Without this check, the underlying LangGraph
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #122, opened 2026-04-28
fix: filter deferred/programmatic tools out of toolSchemaTokens accounting

### LINKED ISSUE #121: calculateInstructionTokens overcounts this.tools, causing false-positive empty_messages preflight rejections at low maxContextTokens
## Summary

`AgentContext.calculateInstructionTokens` reports `toolSchemaTokens` as the sum of *every* tool in `this.tools` regardless of whether they will actually be sent to the model. This inflates the framework's instruction-overhead accounting compared to what `getToolsForBinding` (and therefore the actual prompt) contains. At low `maxContextTokens` values, the resulting overcount triggers spurious `empty_messages` errors with messaging that is technically correct (about the *ceiling*) but false in practice (about the *real* prompt).

## Reproduction

1. Configure an agent with many MCP tools where some have `defer_loading: true` (e.g. Google Workspace MCP, ~100 tools).
2. Set a `maxContextTokens` smaller than the worst-case tool ceiling but large enough for the actual bound subset (e.g. `maxContextTokens: 50000` with 101 registered tools, ~13K of which are bound on turn 1).
3. Send a single short message.

**Expected:** request succeeds; the bound subset of tools fits comfortably within the budget.

**Actual:** request fails preflight with:
```
{"type":"empty_messages","info":"Message pruning removed all messages as none fit in the context window. Tool definitions consume 47705 tokens (99% of instructions) across 101 tools, exceeding maxContextTokens (47500). ..."}
```

The provider's billing for the same convo on a higher cap shows the *real* input was ~13K tokens, not 47K — confirming the prompt would have fit.

## Evidence

Debug log on a fresh conversation, turn 1:
```
[agents:prune] Budget
  maxTokens: 66500,
  pruningBudget: 63175,
  effectiveMax: 15101,
  instructionTokens: 48074,
  toolSchemaTokens: 47705,
  systemMessageTokens: 368,
  messageCount: 1,
  calibratedTotalTokens: 16,
  calibrationRatio: 1,
```

Same conversation, turn 3 (after first round-trip):
```
[agents:prune] Budget
  maxTokens: 66500,
  pruningBudget: 63175,
  effectiveMax: 49674,
  instructionTokens: 13501,
  toolSchemaTokens: 13132,
  systemMessageTokens: 368,
  messageCount: 3,
  calibratedTotalTokens: 2525,
  calibrationRatio: 1,
```

Provider-reported input tokens (Google Vertex via OpenRouter) for the **first** Gemini 3.1 Pro call in the same convo: **13,517 tokens**. That's the authoritative ground truth — it's nowhere near 47K.

## Root cause (verified at source)

[`src/agents/AgentContext.ts:694-735`](https://github.com/danny-avila/agents/blob/main/src/agents/AgentContext.ts) `calculateInstructionTokens`:

```ts
const instanceTools: t.GraphTools = [
  ...((this.tools as t.GenericTool[] | undefined) ?? []),
  ...((this.graphTools as t.GenericTool[] | undefined) ?? []),
];

if (instanceTools.length > 0) {
  for (const tool of instanceTools) {
    // ... unconditionally counts every entry in this.tools
    toolTokens += tokenCounter(new SystemMessage(JSON.stringify(jsonSchema)));
  }
}
```

vs. [`src/agents/AgentContext.ts:1012-1027`](https://github.com/danny-avila/agents/blob/main/src/agents/AgentContext.ts) `getToolsForBinding`:

```ts
getToolsForBinding(): t.GraphTools | undefined {
  if (this.toolDefinitions && this.toolDefinitions.length > 0) {
    return this.getEventDrivenToolsForBinding();
  }
  const filtered =
    !this.tools || !this.toolRegistry
      ? this.tools
      : this.filterToolsForBinding(this.tools);
  // ...
}
```

`filterToolsForBinding` excludes `defer_loading: true` tools that aren't in `discoveredToolNames`, plus tools whose `allowed_callers` don't include `'direct'`. The accounting path skips this filter entirely.

The existing test `excludes deferred-and-undiscovered toolDefinitions from toolSchemaTokens` covers the `toolDefinitions` (event-driven) path, but the same filter is missing on the `this.tools` path that LibreChat hits when MCP tools are loaded as instance tools rather than as deferred definitions.

## Impact

- **At high `maxContextTokens`** (e.g. 200K): invisible — the inflated count is still well below budget.
- **At low caps**: spurious `empty_messages` preflight rejections. Conversations that would have fit get blocked.
- **User-facing error message** quotes "47705 tokens (99% of instructions) across 101 tools" — this is technically the ceiling, but is misleading because the real prompt only contains the bound subset (~13K). Operators reading the error are led to think they need to drastically trim their MCP servers when in reality the prompt fits.

## Proposed fix

Mirror `getToolsForBinding`'s filter inside `calculateInstructionTokens` — skip `defer_loading: true && !discoveredToolNames.has(name)` and skip tools whose `allowed_callers` exclude `'direct'`. Roughly:

```ts
const eligibleTools = instanceTools.filter((tool) => {
  if (!('name' in tool)) return true;
  const def = this.toolRegistry?.get(tool.name as string);
  if (!def) return true;
  const allowedCallers = def.allowed_callers ?? ['direct'];
  if (!allowedCallers.includes('direct')) return false;
  if (def.defer_loading === true && !this.discoveredToolNames.has(tool.name as string)) {
    return false;
  }
  return true;
});
```

Re-running `calculateInstructionTokens` after `markToolsAsDiscovered` would then give a count that converges with what `getToolsForBinding` actually emits.

## Environment

- LibreChat v0.8.5
- @librechat/agents (bundled)
- Provider: Google Gemini 3.1 Pro Preview via OpenRouter custom endpoint
- MCP servers: Google Workspace MCP v1.19.0 (101 tools), plus a few smaller MCPs
- Endpoint mode: ephemeral agent (modelSpec preset → `endpoint: "OpenRouter"`)

danny-avila:
## Summary

Fixes #121.

`AgentContext.calculateInstructionTokens` summed every entry in `this.tools` (and `this.graphTools`) regardless of whether it would be sent to the model, while `getToolsForBinding` filters `this.tools` through `filterToolsForBinding` (excluding `defer_loading: true && !discovered` and `allowed_callers` without `'direct'`). This made `toolSchemaTokens` report the worst-case ceiling instead of what was actually bound.

At low `maxContextTokens` the inflated ceiling triggered spurious `empty_messages` preflight rejections — e.g. `47705 tokens (99% of instructions) across 101 tools` reported as the budget killer when the real prompt was ~13K. Operators reading the error were misled into thinking they had to drastically trim MCP servers.

## Changes

- [src/agents/AgentContext.ts](src/agents/AgentContext.ts): apply `filterToolsForBinding` to `this.tools` inside `calculateInstructionTokens` so accounting matches what `getToolsForBinding` emits. `graphTools` (handoff/subagent) remains unfiltered to mirror the binding path exactly.
- [src/agents/__tests__/AgentContext.test.ts](src/agents/__tests__/AgentContext.test.ts): two regression tests covering the instance-tools path — one for deferred-undiscovered + non-`'direct'` callers being excluded, one for inclusion once discovered via `discoveredTools`. The existing `toolDefinitions` test (`excludes deferred-undiscovered toolDefinitions from toolSchemaTokens`) only covered the event-driven branch and didn't catch this.

## Test plan

- [x] `npx jest src/agents/__tests__/AgentContext.test.ts` — 56/56 pass (54 prior + 2 new)
- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx eslint` on touched files — clean
- [x] Full test suite: only pre-existing live-API failures (openai/anthropic/google/vertexai/cache simple specs, summarization integration) — verified identical on `origin/main`

### DIFF
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -703,9 +703,21 @@ export class AgentContext {
      * populated after `fromConfig()` kicks off the initial calculation, so
      * callers that mutate `graphTools` must re-trigger this method to
      * refresh `toolSchemaTokens`.
+     *
+     * Apply the same registry-based filter that `getToolsForBinding` uses on
+     * `this.tools` so accounting reflects what is actually bound: skip tools
+     * whose `allowed_callers` exclude `'direct'` and tools with
+     * `defer_loading: true` that haven't been discovered yet. Without this,
+     * `toolSchemaTokens` reports the worst-case ceiling (all registered
+     * tools) instead of the bound subset, which can trigger spurious
+     * `empty_messages` preflight rejections at low `maxContextTokens`.
      */
+    const filteredInstanceTools =
+      this.tools && this.toolRegistry
+        ? this.filterToolsForBinding(this.tools)
+        : (this.tools ?? []);
     const instanceTools: t.GraphTools = [
-      ...((this.tools as t.GenericTool[] | undefined) ?? []),
+      ...(filteredInstanceTools as t.GenericTool[]),
       ...((this.graphTools as t.GenericTool[] | undefined) ?? []),
     ];
 
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -919,8 +919,16 @@ export class AgentContext {
    */
   getTokenBudgetBreakdown(messages?: BaseMessage[]): t.TokenBudgetBreakdown {
     const maxContextTokens = this.maxContextTokens ?? 0;
-    const toolCount =
-      (this.tools?.length ?? 0) + this.getActiveToolDefinitions().length;
+    /**
+     * Derive `toolCount` from `getToolsForBinding()` so the diagnostic stays
+     * aligned with what is actually bound to the model — and with what
+     * `calculateInstructionTokens` counts into `toolSchemaTokens`. Using raw
+     * `this.tools.length` would inflate the count whenever the registry
+     * marks instance tools as deferred-undiscovered or non-`'direct'`,
+     * producing the same misleading "N tools" diagnostic this fix is meant
+     * to eliminate.
+     */
+    const toolCount = this.getToolsForBinding()?.length ?? 0;
     const messageCount = messages?.length ?? 0;
 
     let messageTokens = 0;
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -681,10 +681,47 @@ export class AgentContext {
     if (!this.toolDefinitions) {
       return [];
     }
-    return this.toolDefinitions.filter(
-      (def) =>
+    /**
+     * Mirror `getEventDrivenToolsForBinding`'s gate: a definition is only
+     * bound to the model when its `allowed_callers` include `'direct'` and
+     * (if deferred) it has been discovered. Filtering by `defer_loading`
+     * alone left programmatic-only definitions counted in
+     * `toolSchemaTokens` even though they were never bound.
+     */
+    return this.toolDefinitions.filter((def) => {
+      const allowedCallers = def.allowed_callers ?? ['direct'];
+      if (!allowedCallers.includes('direct')) {
+        return false;
+      }
+      return (
         def.defer_loading !== true || this.discoveredToolNames.has(def.name)
-    );
+      );
+    });
+  }
+
+  /**
+   * Single source of truth for "which entries of `this.tools` should be
+   * treated as actually bound". Callers:
+   *   - `getToolsForBinding` (non-event-driven branch)
+   *   - `getEventDrivenToolsForBinding` (appends instance tools alongside
+   *     schema-only definitions)
+   *   - `calculateInstructionTokens` (counts schema bytes for accounting)
+   *
+   * In event-driven mode (`toolDefinitions` present) instance tools are
+   * appended unfiltered; outside event-driven mode they pass through
+   * `filterToolsForBinding`. Centralizing the decision here prevents the
+   * accounting/binding paths from drifting apart, which was the root
+   * cause of the original miscount.
+   */
+  private getEffectiveInstanceTools(): t.GraphTools | undefined {
+    if (!this.tools) {
+      return undefined;
+    }
+    const isEventDriven = (this.toolDefinitions?.length ?? 0) > 0;
+    if (isEventDriven || !this.toolRegistry) {
+      return this.tools;
+    }
+    return this.filterToolsForBinding(this.tools);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Pushed [ed0feb9](https://github.com/danny-avila/agents/pull/122/commits/ed0feb9) addressing Codex's feedback.

**Codex's concern was correct.** `getEventDrivenToolsForBinding` appends `this.tools` UNFILTERED when `toolDefinitions` is present — so unconditionally filtering `this.tools` in the accounting path would have under-counted instance tools that had deferred or non-`'direct'` registry entries (still bound, no longer counted).

The fix now mirrors `getToolsForBinding`'s branching exactly: only run `filterToolsForBinding` on `this.tools` in the non-event-driven path. In event-driven mode, deferred tools are represented via `toolDefinitions` and counted by `getActiveToolDefinitions()`; `this.tools` stays unfiltered to match what gets bound.

New regression test: `does not filter instance tools in event-driven mode (matches getEventDrivenToolsForBinding)` — sets up `toolDefinitions: [activeDef]` plus `tools: [nativeTool]` where `nativeTool` has a deferred-undiscovered registry entry, and asserts the native tool **still increases** `toolSchemaTokens`.

**On the CI failure:** the `summarization.test.ts` › `summary tokenCount matches local token counter` failure is a flaky integration test that hits a real LLM and asserts variance < 0.25 between the model's reported token count and a local counter. This run got 0.27. The test is non-deterministic (different summary text each run) and unrelated to this change — confirmed by inspecting both the test code and the surrounding live-API spec failures. All 57 `AgentContext.test.ts` tests pass locally.

danny-avila:
@codex review

danny-avila:
Pushed [5f96f9f](https://github.com/danny-avila/agents/pull/122/commits/5f96f9f) addressing this.

**Codex's concern was correct.** Aligning `toolSchemaTokens` without aligning `toolCount` half-defeats the fix — the user-facing error at [Graph.ts:987](https://github.com/danny-avila/agents/blob/main/src/graphs/Graph.ts#L987) reads *"Tool definitions consume X tokens across N tools, exceeding maxContextTokens"*, so an accurate X with an inflated N still misleads operators about how many tools are bound.

Fix: derive `toolCount` from `getToolsForBinding()?.length`. That's the same source `calculateInstructionTokens` now mirrors, so `toolSchemaTokens` and `toolCount` stay aligned across all paths (non-event-driven, event-driven, with or without `graphTools`). The existing `toolDefinitions`-path test continues to pass; new test `getTokenBudgetBreakdown toolCount excludes deferred-undiscovered instance tools` covers the instance-tools path that was previously unfiltered.

All 58 `AgentContext.test.ts` tests pass, plus 151 in the broader agents/token-accounting suites. Lint, typecheck, prettier all clean.

danny-avila:
Worked through the comprehensive review. Audit + resolution:

**Finding 1 (toolCount mismatch) — Already resolved.** Fixed in [5f96f9f](https://github.com/danny-avila/agents/pull/122/commits/5f96f9f) (the previous round addressing the Codex P3). The reviewer was looking at a snapshot from before that commit. `getTokenBudgetBreakdown` now derives `toolCount` from `getToolsForBinding()?.length`. No action needed.

**Finding 2 (duplicated branching / drift risk) — Resolved in [628bb18](https://github.com/danny-avila/agents/pull/122/commits/628bb18).** Extracted `getEffectiveInstanceTools()` as the single source of truth for "which entries of `this.tools` are bound to the model". All three call sites now derive from it:
- `getToolsForBinding` (non-event-driven branch)
- `getEventDrivenToolsForBinding` (instance-tool append)
- `calculateInstructionTokens` (accounting)

The event-driven gate and registry-filter decision now live in exactly one place. A future change to either rule updates all paths together. I considered the reviewer's alternative (have `calculateInstructionTokens` iterate `getToolsForBinding()` directly) but rejected it: schema-only tools created by `createSchemaOnlyTool` use `.schema = parameters`, which `toJsonSchema` would emit without the `{ type: 'function', function: {...} }` wrapping that the existing accounting path uses for `toolDefinitions`. Switching paths would silently change per-tool token counts in event-driven mode.

**Finding 3 (JSDoc method-name hyphenation) — Resolved in [628bb18](https://github.com/danny-avila/agents/pull/122/commits/628bb18).** The JSDoc rewrite for the new helper eliminates the `getEventDriven-/ToolsForBinding` split. Verified with grep — no hyphenated method names remain.

**Finding 4 (programmatic-only toolDefinitions over-counted) — Resolved in [628bb18](https://github.com/danny-avila/agents/pull/122/commits/628bb18).** Confirmed real (not theoretical): `code_execution`-only `LCTool` definitions exist throughout the codebase ([`ProgrammaticToolCalling.test.ts`](https://github.com/danny-avila/agents/blob/main/src/tools/__tests__/ProgrammaticToolCalling.test.ts), [`mockTools.ts`](https://github.com/danny-avila/agents/blob/main/src/test/mockTools.ts), etc.). `getActiveToolDefinitions` now filters by `allowed_callers` to match the gate `getEventDrivenToolsForBinding` applies. `getEventDrivenToolsForBinding` itself now reuses `getActiveToolDefinitions()` instead of inlining the same filter — second drift vector eliminated. Added regression test `excludes programmatic-only toolDefinitions from toolSchemaTokens`.

**Verification:**
- 59/59 `AgentContext.test.ts` tests pass
- 235/235 across `src/agents`, prune, token-accounting (pipeline + e2e), subagent, toolHooks, ProgrammaticToolCalling
- Lint, typecheck, prettier all clean

danny-avila:
@codex review

danny-avila:
Follow-up review audit:

**Finding 1 (graphTools toolCount unproven) — Resolved in [19577ba](https://github.com/danny-avila/agents/pull/122/commits/19577ba).** Added `getTokenBudgetBreakdown toolCount includes graphTools` — sets `tools: [direct_tool]` and assigns `graphTools = [handoff_tool]`, asserts `toolCount === 2`. Locks in the behavioral improvement (graphTools now aligned with toolSchemaTokens accounting).

**Finding 2 (allocation in event-driven mode) — Skipping per reviewer's own note** ("This is a micro-optimization that's not worth introducing now"). All callers are confirmed diagnostic-only paths.

**Finding 3 (overlapping name double-count) — Skipping.** Pre-existing, not introduced here, marked "for completeness only" by the reviewer. A misconfiguration scenario unlikely in practice.

**Verification:** 60/60 `AgentContext.test.ts` tests pass, lint and prettier clean.

PR #124, opened 2026-04-30
fix(anthropic): drop whitespace-only text blocks before send

danny-avila:
## Summary

Fixes [LibreChat discussion #12806](https://github.com/danny-avila/LibreChat/discussions/12806) — Claude + web_search + a follow-up turn fails with:

```
messages: text content blocks must contain non-whitespace text
```

## Root cause

\`_formatContent\` in \`src/llm/anthropic/utils/message_inputs.ts\` filters empty text blocks before sending the payload to the Anthropic API, but the check only catches strict \`text === ''\`. Blocks with whitespace-only text (\`' '\`, \`'\n'\`, \`'\t'\`, etc.) are passed through. The Anthropic API rejects those when the message has no other non-whitespace text content.

I confirmed the API rules by hitting \`messages.create\` directly:

| Content shape | API result |
|---|---|
| \`[text:'']\` (anywhere) | rejected — \"text content blocks must be non-empty\" |
| \`[text:' ']\` (only block) | rejected — \"text content blocks must contain non-whitespace text\" |
| \`[text:' ', text:'real']\` | accepted |
| \`[text:'\n']\` (only block) | rejected — \"text content blocks must contain non-whitespace text\" |

The reporter's exact error message corresponds to row 2/4 — Anthropic returned a web_search response whose only text block was whitespace, so re-sending the history triggers the rejection.

## The fix

Broaden the existing filter to drop any text block whose trimmed content is empty. One-liner change in \`message_inputs.ts:611\`:

\`\`\`diff
-        !(block.type === 'text' && 'text' in block && block.text === '')
+        !(
+          block.type === 'text' &&
+          'text' in block &&
+          typeof block.text === 'string' &&
+          block.text.trim() === ''
+        )
\`\`\`

Whitespace-only text blocks have no semantic value — Anthropic's web_search responses can include them as artifacts between \`server_tool_use\` and \`web_search_tool_result\` blocks. Dropping them is safe and preserves all citation-relevant blocks (per the existing comment at lines 568–574).

## Why the reporter saw it gated by Custom Instructions OR cache OFF

The condition for the bug to manifest is: the assistant message that's about to be re-sent contains only whitespace text alongside its tool blocks. The reporter's environment hits this. The \`addCacheControl\` path at \`graphs/Graph.ts:879\`–\`888\` is gated to run only when \`promptCache === true\` AND \`!systemRunnable\`, which probably correlates with whatever response shape avoids the trigger in their setup. The fix resolves the failure mode regardless of cache state or system prompt.

## Test plan

- [x] New regression cases in \`server-tool-inputs.test.ts\` (5 whitespace variants: space, newline, tab, multiple spaces, mixed) all pass with the fix and previously failed.
- [x] Existing \`'filters out empty text blocks from array content'\` test (strict empty case) still passes — no regression on the original behavior.
- [x] All 11 cases in \`server-tool-inputs.test.ts\` green: \`npx jest src/llm/anthropic/utils/server-tool-inputs.test.ts\`.
- [x] Live e2e against \`claude-opus-4-7\`, \`claude-sonnet-4-6\`, \`claude-haiku-4-5\`, \`claude-opus-4-1-20250805\`, \`claude-sonnet-4-20250514\` — turn 1 and turn 2 both succeed with the fix in place.
- [x] Direct API probes confirm the rejection rules above.

### DIFF
--- a/src/llm/anthropic/utils/message_inputs.ts
+++ b/src/llm/anthropic/utils/message_inputs.ts
@@ -611,7 +611,12 @@ function _formatContent(message: BaseMessage) {
     return contentBlocks.filter(
       (block) =>
         block !== null &&
-        !(block.type === 'text' && 'text' in block && block.text === '')
+        !(
+          block.type === 'text' &&
+          'text' in block &&
+          typeof block.text === 'string' &&
+          block.text.trim() === ''
+        )
     );
   }
 }
--- a/src/llm/anthropic/utils/server-tool-inputs.test.ts
+++ b/src/llm/anthropic/utils/server-tool-inputs.test.ts
@@ -346,4 +346,81 @@ describe('_convertMessagesToAnthropicPayload — server tool use (web search) mu
     expect(textBlocks).toHaveLength(1);
     expect(textBlocks[0].text).toBe('Here are the results.');
   });
+
+  /**
+   * Regression for LibreChat discussion #12806.
+   *
+   * Anthropic web_search responses can include text blocks whose text is
+   * whitespace-only (e.g. ' ', '\n', '\t') alongside server_tool_use and
+   * web_search_tool_result blocks. On follow-up turns the API rejects these
+   * with: "messages: text content blocks must contain non-whitespace text".
+   *
+   * The empty-string check alone is insufficient — the filter must drop any
+   * text block whose trimmed content is empty.
+   */
+  it.each([
+    ['single space', ' '],
+    ['newline', '\n'],
+    ['tab', '\t'],
+    ['multiple spaces', '   '],
+    ['mixed whitespace', ' \n\t '],
+  ])(
+    'filters whitespace-only text blocks from array content (%s)',
+    (_label, whitespace) => {
+      const messageHistory: BaseMessage[] = [
+        new HumanMessage('search for X'),
+        new AIMessage({
+          content: [
+            { type: 'text', text: whitespace },
+            {
+              type: 'server_tool_use',
+              id: 'srvtoolu_1',
+              name: 'web_search',
+              input: { query: 'X' },
+            },
+            {
+              type: 'web_search_tool_result',
+              tool_use_id: 'srvtoolu_1',
+              content: [
+                {
+                  type: 'web_search_result',
+                  url: 'https://example.com',
+                  title: 'Result',
+                  encrypted_content: 'abc',
+                  page_age: '1d',
+                },
+              ],
+            },
+            { type: 'text', text: 'Here are the results.' },
+          ],
+          tool_calls: [
+            {
+              id: 'srvtoolu_1',
+              name: 'web_search',
+              args: { query: 'X' },
+              type: 'tool_call',
+            },
+          ],
+        }),
+        new HumanMessage('follow up'),
+      ];
+
+      const { messages } = _convertMessagesToAnthropicPayload(messageHistory);
+      const assistantContent = messages[1].content as any[];
+
+      const whitespaceTextBlocks = assistantContent.filter(
+        (b: any) =>
+          b.type === 'text' &&
+          typeof b.text === 'string' &&
+          b.text.trim() === ''
+      );
+      expect(whitespaceTextBlocks).toHaveLength(0);
+
+      const textBlocks = assistantContent.filter(
+        (b: any) => b.type === 'text'
+      );
+      expect(textBlocks).toHaveLength(1);
+      expect(textBlocks[0].text).toBe('Here are the results.');
+    }
+  );
 });
--- a/src/specs/anthropic.simple.test.ts
+++ b/src/specs/anthropic.simple.test.ts
@@ -376,6 +376,67 @@ describe(`${capitalizeFirstLetter(provider)} Streaming Tests`, () => {
     );
   });
 
+  test(`${capitalizeFirstLetter(provider)}: follow-up after assistant message with only whitespace text content`, async () => {
+    /**
+     * Regression for LibreChat discussion #12806.
+     *
+     * The Anthropic API has two distinct rejection rules (verified against
+     * the live API):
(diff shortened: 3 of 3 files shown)


danny-avila:
Added a real-API regression test in `src/specs/anthropic.simple.test.ts` (the existing real-API spec file) that runs against the actual Anthropic API and exhibits the exact reporter error.

Verified the test is a valid regression:

- **With the fix:** PASS in 5.2s
- **Without the fix:** FAIL with `400 messages: text content blocks must contain non-whitespace text` — the verbatim error from #12806

```
× Anthropic: follow-up after assistant message with only whitespace text content (254 ms)
   400 {"type":"error","error":{"type":"invalid_request_error","message":"messages: text content blocks must contain non-whitespace text"},...}
```

The test uses an assistant message with only `[{type:'text', text:' '}]` content (no tools) — the precise shape that triggers Anthropic's rule. With tool blocks present (e.g. `server_tool_use` from web_search) the API is more lenient and accepts whitespace text alongside, so a tool-flavored repro can't be made deterministic. The minimal-shape test catches the underlying filter gap regardless of which upstream code path produces such a message.

I also confirmed via direct `messages.create` probes (no SDK in between) the exact rejection rules:

| Shape | API |
|---|---|
| `[text:'']` (anywhere) | rejected — "must be non-empty" |
| `[text:' ']` (only block) | rejected — "must contain non-whitespace text" |
| `[text:' ', text:'real']` | accepted |
| `[text:' ', server_tool_use, ...]` | accepted (tool blocks make it lenient) |

So the existing strict `text === ''` filter caught rule 1 but missed rule 2 — that's the gap.

danny-avila:
@codex review

danny-avila:
### Why this fix exists — progression of related work

The two filters that should have caught this drifted out of sync over a couple of months. This PR aligns them.

**Feb 21, 2026 — `b253810`** *(direct commit to `main`, not a PR)*
> `fix: message formatting to strip reasoning and thinking blocks, and handle whitespace-only text in agent messages`

Added the `else { if (part.type === ContentTypes.TEXT && !String(part.text ?? '').trim()) continue; }` filter in [`src/messages/format.ts`](https://github.com/danny-avila/agents/blob/main/src/messages/format.ts) inside `formatAssistantMessage`. This guards the **DB → BaseMessage** path in `formatAgentMessages`, which LibreChat hits when reloading conversation history between turns. It correctly used `.trim()` — catches both empty and whitespace-only text.

**Apr 13, 2026 — [PR #92](https://github.com/danny-avila/agents/pull/92)** *(`🔍 fix: Preserve Anthropic server_tool_use block types`, squashed into `360bafa`)*
> commit `4b888cd`: `fix: filter empty text blocks and fix empty content fallback`
>
> *"Filter out empty text blocks ({type:'text', text:''}) from `_formatContent` output — these can appear when the model emits a content_block_start with empty text before server tool calls"*

Added a second-line-of-defense filter in [`src/llm/anthropic/utils/message_inputs.ts`](https://github.com/danny-avila/agents/blob/main/src/llm/anthropic/utils/message_inputs.ts) inside `_formatContent`. This is the actual wire-send conversion (called from `_streamResponseChunks`). The filter targeted exactly what was observed empirically — the strict-empty `text: ''` block — and used `block.text === ''`. **Whitespace-only blocks (`' '`, `'\n'`, `'\t'`) weren't on the radar**, partly because Anthropic returns a *different* error message for them:

| Shape | API rejection (verified live) |
|---|---|
| `text: ''` (anywhere) | `"text content blocks must be non-empty"` |
| `text: ' '` / `'\n'` (sole content) | `"text content blocks must contain non-whitespace text"` |

PR #92 saw the first error class only.

**Apr 24, 2026 — [LibreChat discussion #12806](https://github.com/danny-avila/LibreChat/discussions/12806)**
Reporter hits the second error class consistently — when web_search runs and *either* Custom Instructions are set *or* prompt caching is off (i.e. when [`addCacheControl` is gated off in `Graph.ts:879-888`](https://github.com/danny-avila/agents/blob/main/src/graphs/Graph.ts)). The format.ts filter from `b253810` doesn't help here because:

- It only runs on the `formatAgentMessages` (DB-reload) path. In-memory BaseMessages within a single `Run` (mid-iteration) bypass it.
- It only catches `part.type === ContentTypes.TEXT`. If the persisted shape preserves a streaming variant (e.g. `text_delta`), the catch-all `else` skips the trim check.

So whitespace text reaches `_formatContent` (the wire-send) — and PR #92's strict-empty filter lets it through.

### What this PR does

Bring `_formatContent` to parity with `formatAssistantMessage`: both now use `.trim() === ''`. One change, two-layer alignment.

```diff
-  !(block.type === 'text' && 'text' in block && block.text === '')
+  !(
+    block.type === 'text' &&
+    'text' in block &&
+    typeof block.text === 'string' &&
+    block.text.trim() === ''
+  )
```

### What was missed before

- **PR #92** picked the narrowest predicate that fixed the observed symptom rather than mirroring the predicate in `b253810`. If both filters had used `.trim()` from the start, this discussion wouldn't exist.
- There was no real-API regression test exercising the wire-send path. `anthropic.simple.test.ts` covered single-turn web_search and multi-turn-without-web_search, but not the bug pattern (assistant content the API will reject on follow-up). This PR adds one — see prior comment for verification that it FAILS without the fix and PASSES with it.

### Other related history (not directly fixing this, but in the same area)

- `f507a1b`, `360bafa` — prior server_tool_use block-type preservation (also part of #92)
- `fbe896e` — earlier Anthropic web_search edge-case + type validation
- `1b7dad3` — original feat adding Anthropic web_search support
- LibreChat-side: [PR #12631 (Apr 12)](https://github.com/danny-avila/LibreChat/pull/12631) "Scope Web Search Results to Own Turn" — frontend display only, unrelated to the API rejection

PR #147, opened 2026-05-05
🧬 feat: Inherit Parent Configurable Into Subagent Workflow

danny-avila:
## Summary

- `SubagentExecutor` previously invoked the child workflow with a fresh `configurable: { thread_id: childRunId }`, dropping every host-set field the parent had (e.g. `requestBody`, `user`, `userMCPAuthMap`).
- Subagent tool dispatches to the parent's `ON_TOOL_EXECUTE` handler arrived with that stripped view, so hosts that read context from the configurable (e.g. LibreChat's MCP body-placeholder substitution and per-user connection lookup) silently failed in subagent contexts even though the same handler runs in the same process.
- Adds optional `SubagentExecuteParams.parentConfigurable`. When provided, the executor spreads it into the child's `workflow.invoke` configurable, scrubbing run-identity fields (`thread_id` / `run_id` / `parent_run_id`) and overriding `thread_id` with the new `childRunId`. Anything else passes through — the SDK doesn't interpret host-defined keys.
- `Graph.ts` spawn-subagent tool now passes `parentConfigurable: config.configurable` (one-line wire-up; the parent's runnable config is in scope at the call site).
- Bumps to `v3.1.78` so LibreChat can pin against a published, non-dev version.

## Files changed

| File | Note |
|---|---|
| `src/tools/subagent/SubagentExecutor.ts` | Adds `parentConfigurable?: Record<string, unknown>` to `SubagentExecuteParams`; spreads it into the child's invoke configurable with run-identity fields scrubbed and `thread_id` overridden. |
| `src/graphs/Graph.ts` | Spawn-subagent tool passes `parentConfigurable: config.configurable` to `executor.execute`. |
| `src/tools/__tests__/SubagentExecutor.test.ts` | New `parentConfigurable inheritance` describe block — 4 tests: forwarding host fields, overriding `thread_id`, scrubbing run-identity fields, back-compat (`parentConfigurable` omitted). |
| `src/scripts/subagent-configurable-inheritance.ts` | Live verification script — spawns a self-subagent, asserts every host-set key on parent's outer configurable shows up on subagent `ON_TOOL_EXECUTE` dispatches with `thread_id` overridden. |
| `package.json` | `3.1.78-dev.0` → `3.1.78`. |

## Back-compat

`parentConfigurable` is optional. Hosts that don't pass it get the old isolated behavior unchanged.

## Test plan

- [x] `npx jest --testPathPatterns=SubagentExecutor.test` — 58/58 pass (54 prior + 4 new).
- [x] `bun -r dotenv/config ./src/scripts/subagent-configurable-inheritance.ts` against `gpt-4o-mini` — passes; subagent's `ON_TOOL_EXECUTE` sees `user`, `user_id`, `requestBody`, `userMCPAuthMap` from parent and `thread_id` is the new childRunId.
- [ ] Downstream verification in LibreChat once `@librechat/agents@3.1.78` is published.

## Downstream

LibreChat will bump `@librechat/agents` to `3.1.78` in a follow-up PR. The user/user_id injection added in [danny-avila/LibreChat#12950](https://github.com/danny-avila/LibreChat/pull/12950) becomes redundant with this change but stays as defense-in-depth.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@librechat/agents",
-  "version": "3.1.78-dev.0",
+  "version": "3.1.78",
   "main": "./dist/cjs/main.cjs",
   "module": "./dist/esm/main.mjs",
   "types": "./dist/types/index.d.ts",
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -1471,6 +1471,15 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
             subagentType,
             threadId,
             parentToolCallId,
+            /**
+             * Forward the parent's `configurable` so host-set fields
+             * (`requestBody`, `user`, etc.) propagate into the child
+             * workflow. The executor scrubs run-identity fields before
+             * forwarding — see `SubagentExecuteParams.parentConfigurable`.
+             */
+            parentConfigurable: config.configurable as
+              | Record<string, unknown>
+              | undefined,
           });
           return result.content;
         }, buildSubagentToolParams(resolvedConfigs));
--- a/src/scripts/subagent-configurable-inheritance.ts
+++ b/src/scripts/subagent-configurable-inheritance.ts
@@ -1,0 +1,208 @@
+import { config } from 'dotenv';
+config();
+
+import { HumanMessage } from '@langchain/core/messages';
+import type * as t from '@/types';
+import { ChatModelStreamHandler } from '@/stream';
+import { ToolEndHandler, ModelEndHandler } from '@/events';
+import { Providers, GraphEvents } from '@/common';
+import { Run } from '@/run';
+
+/**
+ * Live verification that host-set fields on the parent's outer
+ * `configurable` (e.g. `requestBody`, `user`, `userMCPAuthMap`)
+ * propagate into the subagent's `ON_TOOL_EXECUTE` dispatches.
+ *
+ * Pass criteria: when the SUBAGENT calls the calculator tool, the
+ * `data.configurable` arriving at the parent's ON_TOOL_EXECUTE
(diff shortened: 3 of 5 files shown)


danny-avila:
Pushed e75c0dc addressing the run-identity scrubbing question.

**Reviewer point:** the subagent IS the same run/conversation user-perceptually, and `SubagentExecutor` already commits to that with `sessionId: this.parentRunId` for `SubagentStart`/`SubagentStop`. Scrubbing `run_id`/`parent_run_id`/`thread_id` contradicted that contract.

**What changed:** dropped all three deletes. `parentConfigurable` is now forwarded verbatim. `thread_id` only falls back to `childRunId` if parent didn't supply one.

**Empirical finding from the live script (gpt-4o + LangGraph runtime):**
- `thread_id`: propagates ✅
- `parent_run_id`: propagates when host supplies it ✅
- host-set keys (`user`, `requestBody`, `userMCPAuthMap`, etc.): all propagate ✅
- `run_id`: **LangGraph runtime overwrites** at child invoke time regardless of what we forward. The SDK forwards correctly, but tools downstream see the runtime-assigned runId, not parent's. Hosts that need parent-scoped run identity should use a host-defined key (e.g. `requestBody.messageId`) instead of `run_id`. Doc updated to call this out.

**Test changes:**
- replaced "scrubs run-identity fields" → "forwards run-identity fields verbatim into child invoke configurable" (with the LangGraph-runtime override note)
- added "inherits parent thread_id when supplied" and "falls back to childRunId when not"
- live script switched to non-self subagent (math-worker with calculator; parent has no tools) to force the spawn path and reliably distinguish parent vs subagent dispatches by `agentId`

Future work flagged in the doc: per-spawn-type inheritance config (e.g. background/async subagents may want isolation).

PR #140, opened 2026-05-04
fix: Cross-provider tool-call ID + thinking-block compatibility

danny-avila:
## Summary

Two narrowly-scoped fixes for cross-provider 400 errors observed when conversation history crosses provider boundaries. Both bugs were confirmed live against real APIs before fixing — see `src/scripts/poc-live-x-provider.ts` for the reproducer.

### Fix 1 — Tool-call IDs survive an OpenAI Responses → Anthropic handoff

OpenAI Responses-style tool-call IDs (e.g. `fc_...|call_...`) routinely contain `|` and exceed 64 chars. Anthropic enforces `^[a-zA-Z0-9_-]+$` and a 64-char cap and rejects with 400 on replay.

Added `normalizeAnthropicToolCallId` (pure, deterministic) and applied at the four wire-bound sites in `src/llm/anthropic/utils/message_inputs.ts`:
- `_convertLangChainToolCallToAnthropic` (assistant tool_use.id)
- the three `tool_result.tool_use_id` constructions in `_ensureMessageContents`

Server tool IDs (`srvtoolu_` prefix) are left untouched. Anthropic-native IDs that already comply pass through unchanged. Determinism means paired `tool_use.id` and `tool_result.tool_use_id` stay matched without any session-scoped Map.

### Fix 2 — Anthropic thinking blocks survive a → OpenAI handoff

`thinking` and `redacted_thinking` content blocks were being forwarded verbatim to OpenAI Chat Completions, which 400s with *Invalid value: 'thinking'*. Even when accepted (server may strip), the reasoning narrative is invisible to the next model.

`_convertMessagesToOpenAIParams` now:
- flattens non-empty `thinking` blocks to `<thinking>...</thinking>` text so the reasoning narrative survives as in-band context
- drops empty `thinking` blocks (empty content is rejected by some providers)
- drops `redacted_thinking` entirely (encrypted, useless to non-Anthropic models)
- continues to pass thinking blocks through verbatim for Claude served via OpenAI-shaped surfaces (e.g. OpenRouter), detected by model name

## Test plan

Unit tests added — all passing:
- [x] `src/llm/anthropic/utils/tool-id-normalization.test.ts` — 9 tests covering helper + round-trip through `_convertMessagesToAnthropicPayload`
- [x] `src/llm/openai/utils/messages.test.ts` — 5 new test cases under "cross-provider thinking block handling"
[...]

PR #153, opened 2026-05-07
fix: Normalize tool-call IDs for Anthropic compatibility

danny-avila:
## Summary

Fixes a cross-provider 400 from Anthropic when tool-call IDs originating from OpenAI's Responses API (or any other ID source that violates Anthropic's regex) flow into an Anthropic call as part of replayed history.

## The bug

OpenAI Responses API generates tool-call IDs that can:
- Exceed 64 characters
- Contain `|` and other characters outside Anthropic's `^[a-zA-Z0-9_-]+$` constraint

When this library replays such history to Anthropic (e.g. for an OpenAI → Anthropic handoff), the IDs flow through verbatim and Anthropic 400s with:

```
messages.1.content.1.tool_use.id: String should match pattern '^[a-zA-Z0-9_-]+$'
```

## The fix

`normalizeAnthropicToolCallId` is applied at the four wire-bound sites in `src/llm/anthropic/utils/message_inputs.ts`:

- `_convertLangChainToolCallToAnthropic` — assistant `tool_use.id`
- The three `tool_result.tool_use_id` constructions in `_ensureMessageContents`

Compliant IDs pass through unchanged. Non-compliant IDs are sanitized (`[^a-zA-Z0-9_-]` → `_`), then truncated to 53 chars and suffixed with `_<10-hex-char SHA-256>` of the **original** input. This ensures:

1. Output always satisfies Anthropic's regex and 64-char limit
2. Two long IDs that share a 64-char prefix produce **distinct** outputs (preventing "tool_use ids must be unique" rejections)
3. Two short IDs differing only by an invalid char (e.g. `a|b` vs `a.b`) also produce distinct outputs
4. The function is pure and deterministic — paired `tool_use.id` and `tool_result.tool_use_id` stay matched without any session map

Server tool IDs (`srvtoolu_` prefix) are explicitly excluded — they're Anthropic-internal and always compliant.

## Test plan

- [x] 12 unit tests in `src/llm/anthropic/utils/tool-id-normalization.test.ts`:
  - Helper-level: compliant pass-through, sanitization + hash suffix, length cap, collision-resistance for shared-prefix IDs, short-ID disambiguation, determinism, undefined/empty edge cases
  - Integration: round-trip through `_convertMessagesToAnthropicPayload` confirms `tool_use.id` and `tool_result.tool_use_id` receive the same normalized value
- [x] Full `src/llm/anthropic` suite still passes (1 pre-existing flake in abort-signal test, reproduces on baseline)
- [x] Live verified against the Anthropic API: previously-rejected payload with `fc_67abc1234def567|call_abc123def456ghi789jkl0mnopqrs` now succeeds. Collision case with two 96-char IDs sharing an 80-char prefix also accepted.

## Scope

Intentionally narrow. Related cross-provider concerns (Anthropic thinking-block flatten for OpenAI, wrapper coverage) are tracked separately on #140.

### DIFF
--- a/src/llm/anthropic/utils/message_inputs.ts
+++ b/src/llm/anthropic/utils/message_inputs.ts
@@ -3,6 +3,7 @@
 /**
  * This util file contains functions for converting LangChain messages to Anthropic messages.
  */
+import { createHash } from 'node:crypto';
 import {
   type BaseMessage,
   type SystemMessage,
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #159, opened 2026-05-08
🪪 fix: align thought-signature dispatch by position to fix Vertex tool round-trip

danny-avila:
## Summary

`fixThoughtSignatures` correlates each AI message with a Gemini "model" content block to re-attach `thoughtSignature` onto functionCall parts. Pre-fix it filtered AI messages to only those with non-empty signatures, then walked the *filtered* array against the *unfiltered* modelContents — as soon as any prior turn's AI message lacked signatures, the indices desynced and the toolcall AI message was matched against the wrong content block. Real signature dropped, Vertex returns 400 on the follow-up call.

Reported in [danny-avila/LibreChat#13006](https://github.com/danny-avila/LibreChat/issues/13006). Triggers reliably on `gemini-3.1-flash-lite-preview` whenever the conversation has ANY plain-text assistant turn before a tool-using turn.

## Live repro

```
turn 1: HumanMessage("hi"), AIMessage("Hello!"),       ← no signatures
turn 2: HumanMessage("run X"), AIMessage(toolcall),    ← has signatures
        ToolMessage(result)
```

Old logic:
- `aiMessages.filter(has signatures)` → [toolcall AI] (1 entry)
- `modelContents` → ["Hello!", functionCall] (2 entries)
- `count = min(1, 2) = 1`
- iteration 0: pairs `toolcall AI` with `"Hello!"` content (NO functionCall part to attach to)
- functionCall content never visited → no signature attached

Vertex 400 confirmed live. With the fix (walk all AI messages by position, skip non-signature ones inline), signature lands on the right functionCall and the round-trip completes.

## Changes

- `src/llm/vertexai/index.ts` — drop the `has signatures` pre-filter, skip inline so positional alignment is preserved.
- `src/llm/vertexai/fixThoughtSignatures.test.ts` — new file, 5 tests covering the failure case + multi-turn dispatch + idempotence + empty-signature handling.
- Function exported so it can be tested directly.

## Verification

- Unit tests: `npx jest fixThoughtSignatures.test.ts` → 5 passed.
- Live integration: `gemini-3.1-flash-lite-preview` Vertex stream + bash-style tool round-trip. Without fix: `Google request failed with status code 400`. With fix: response completes.

## Test plan

- [x] Repro confirmed before fix (400 from Vertex)
- [x] Repro confirmed fixed after the change
- [x] Unit tests pass
- [x] No change to behavior when prior turns also have signatures (existing case path)

### DIFF
--- a/src/llm/vertexai/fixThoughtSignatures.test.ts
+++ b/src/llm/vertexai/fixThoughtSignatures.test.ts
@@ -1,0 +1,154 @@
+import { expect, test, describe } from '@jest/globals';
+import type { GeminiContent } from '@langchain/google-common';
+import { AIMessage, HumanMessage, ToolMessage } from '@langchain/core/messages';
+import { fixThoughtSignatures } from './index';
+
+const SIG_A = 'AY89a1/sigA==';
+const SIG_B = 'AY89a1/sigB==';
+
+const buildContents = (
+  blocks: Array<['user' | 'model' | 'function', GeminiContent['parts']]>
+): GeminiContent[] =>
+  blocks.map(([role, parts]) => ({ role, parts }) as GeminiContent);
+
+describe('fixThoughtSignatures', () => {
+  test('attaches signature to functionCall part when prior turn is a plain-text AI message (issue LibreChat#13006-followup)', () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #220, opened 2026-06-05
feat: optional deterministic Langfuse trace IDs (seed from runId)

graphaelli:
## What

Adds an opt-in `LangfuseConfig.deterministicTraceId`. When enabled, a run's
**root** Langfuse trace id is derived from its `runId` as `sha256(runId)` → first
32 hex chars (matching `@langfuse/tracing`'s `createTraceId`), instead of a
random id. Off by default — fully backwards compatible.

## Why

External systems can't attach a Langfuse **score** or observation to a run's
trace after the fact (e.g. user 👍/👎 feedback submitted later from the host
app) because the trace id is random — they'd need a lookup. With this flag, the
host regenerates the same id from the run/message id and attaches scores with
zero lookups.

This can't be done cleanly from the app layer: injecting a parent span context
from outside makes the run's span a *child*, so it stops being the trace root
and the trace ingests **unnamed**. The trace id must be set where the
OpenTelemetry SDK is configured — here.

## How

- A custom OTEL `IdGenerator` on the Langfuse `BasicTracerProvider` returns the
  seeded id when a per-run seed is present in an `AsyncLocalStorage`, otherwise a
  random id.
- `Run.processStream` wraps stream execution in `runWithTraceIdSeed(runId, …)`
  only when `deterministicTraceId` is set. Title-generation tracing is left
  untouched.

## Tests

New `src/specs/deterministic-trace-id.test.ts` (seeded → deterministic, unseeded
→ random, undefined-seed passthrough). Existing Langfuse specs unchanged and
green; `npm run build` + eslint clean.

## Context

Companion to danny-avila/LibreChat#13537 — surfacing message feedback as Langfuse
scores, which sets this flag and scores by `createTraceId(messageId)`.

---

### Companion feature working end-to-end

The LibreChat change (danny-avila/LibreChat#13537) sets `deterministicTraceId`
and scores by `createTraceId(messageId)`. A 👍/👎 in the chat UI lands as a
Langfuse **score attached to that message's trace** — no lookup:

Langfuse Scores UI showing a user-feedback BOOLEAN score (thumbsUp / clear_well_written)

### DIFF
--- a/src/instrumentation.ts
+++ b/src/instrumentation.ts
@@ -12,7 +12,10 @@ import {
   getContextLangfuseConfig,
 } from '@/langfuseToolOutputTracing';
 import { isPresent } from '@/utils/misc';
+import { AsyncLocalStorage } from 'node:async_hooks';
+import { createHash, randomBytes } from 'node:crypto';
 import type {
+  IdGenerator,
   ReadableSpan,
   Span,
   SpanProcessor,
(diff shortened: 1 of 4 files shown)


graphaelli:
tested end to end with https://github.com/danny-avila/LibreChat/compare/main...graphaelli:LibreChat:feat/langfuse-feedback-scores

danny-avila:
thank you Gil!

danny-avila:
@codex review

danny-avila (approved):

PR #225, opened 2026-06-06
feat(hooks): add executingAgentId to attribute tool batches to their agent

danny-avila:
## Problem

The hook `agentId` is a **subagent-scope marker** — it's `undefined` for top-level agents (documented on `BaseHookInput`, asserted in `subagentHooks.test.ts`). In a **multi-agent graph**, an agent whose tools run via the **traditional `ToolNode` path** (engine/runtime tools, i.e. no `toolDefinitions`) emits `PostToolBatch` / `PreToolUse` / `PostToolUse` / `PermissionDenied` with **no way to attribute the batch to the owning agent**.

Root inconsistency in `Graph.initializeTools`: the **event-driven** path passes `agentId: agentContext?.agentId`, but the **traditional** path **omits it**. So the same agent reports its id when using event-driven tools and `undefined` when using engine/runtime tools. "Fixing" that by populating `agentId` on the traditional path would change the documented subagent-scope semantics and break `subagentHooks.test.ts`.

## Fix (additive, non-breaking)

Introduce **`executingAgentId`** on `BaseHookInput`, always set to the owning agent's id — including top-level agents in a multi-agent graph. `agentId` is left **completely unchanged**, so the subagent-scope contract and all existing hosts are untouched.

- `BaseHookInput.executingAgentId` + `ToolNodeOptions.executingAgentId` (types)
- `ToolNode`: new field + ctor param (defaults to `executingAgentId ?? agentId` so direct callers get it for free), surfaced at all 9 hook-emission sites
- `Graph.initializeTools`: pass `executingAgentId: agentContext?.agentId` on **both** ToolNode paths
- Programmatic bridge: threaded through `ProgrammaticHookContext` → `applyPreToolUseHooksForBridge` so inner programmatic tool calls (`write_file`/`edit_file`) are attributed too

## Verification

- `tsc --noEmit` clean; full `npm run build` clean; `eslint` clean.
- Tool/hook suites pass, incl. new tests: top-level agent has `agentId: undefined` **but** `executingAgentId` set (through the real `Run` pipeline); direct-ToolNode default + override; programmatic-bridge threading.

## Motivation

Found while auditing hook agent-attribution for a multi-agent triage graph. Per-agent hook logic (budgeting/inspecting one agent's tool batches) wasn't reliable for agents on the traditional/engine-tool path, where `agentId` arrives `undefined`. `executingAgentId` gives hosts a dependable "which agent ran this" signal without disturbing the subagent-scope meaning of `agentId`.

### DIFF
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -1218,6 +1218,7 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
         sessions: this.sessions,
         toolDefinitions: toolDefMap,
         agentId: agentContext?.agentId,
+        executingAgentId: agentContext?.agentId,
         toolCallStepIds: this.toolCallStepIds,
         toolRegistry: agentContext?.toolRegistry,
         hookRegistry: this.hookRegistry,
@@ -1262,6 +1263,10 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
       trace: traceToolNode,
       runLangfuse: this.langfuse,
       agentLangfuse: agentContext?.langfuse,
+      // `agentId` is intentionally left unset on this path (it is the
+      // subagent-scope marker); `executingAgentId` always identifies the owning
+      // agent so hooks can attribute the batch even at the top level.
+      executingAgentId: agentContext?.agentId,
       toolCallStepIds: this.toolCallStepIds,
       errorHandler: (data, metadata): Promise<void> =>
         StandardGraph.handleToolCallErrorStatic(this, data, metadata),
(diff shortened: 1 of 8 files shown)


danny-avila:
@codex review

danny-avila:
Addressed both P2 findings in `4f2ad3e`:

1. **Programmatic bridge** — threaded `executingAgentId` through `ProgrammaticHookContext` → `hookContext` → `applyPreToolUseHooksForBridge`, so inner programmatic tool calls (e.g. `write_file`/`edit_file` via the bridge) attribute `PreToolUse` to the owning agent instead of `undefined`. Added a unit test.
2. **Direct ToolNode callers** — default `executingAgentId` to `executingAgentId ?? agentId` so callers who only pass the existing `agentId` option still get attribution. Added a test for the default + explicit override.

104 tool/hook tests pass; `tsc` + `eslint` clean.

danny-avila:
@codex review

PR #228, opened 2026-06-07
fix: honor proxy env vars in web search HTTP calls

### LINKED ISSUE #13483: Web Search: Tavily/Jina/Cohere HTTP calls via axios do not respect HTTP_PROXY/HTTPS_PROXY environment variables
## Problem

The web search feature's HTTP calls (Tavily search, Tavily extract, Jina reranker, Cohere reranker) use `axios` without proxy configuration. In corporate/enterprise environments where all outbound internet traffic must pass through a forward proxy, these calls fail with `502 Bad Gateway` because axios does not honor `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` environment variables by default.

### Environment

- LibreChat deployed in Kubernetes behind a corporate forward proxy
- All outbound internet access requires proxy (`HTTP_PROXY=http://proxy:8080`)
- `PROXY`, `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` env vars are set in the pod
- MCP connections work correctly (they use `undici` with `ProxyAgent` and respect `NO_PROXY`)
- The older Tavily agent tool (`api/app/clients/tools/structured/TavilySearch.js`) works correctly (uses `undici` with `ProxyAgent`)

### Affected Code

`@librechat/agents/src/tools/search/tavily-search.ts`:

```typescript
const response = await axios.post<t.TavilySearchResponse>(
  config.apiUrl,
  payload,
  {
    headers: {
      Authorization: `Bearer ${config.apiKey}`,
      'Content-Type': 'application/json',
    },
    timeout: config.timeout,
  }
);
```

Similarly affected (likely): Tavily extract client, Jina reranker client, Cohere reranker client — any HTTP call in the web search pipeline using `axios`.

### Expected Behavior

Web search HTTP calls should respect standard proxy environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, and LibreChat's own `PROXY` variable), consistent with how MCP connections handle proxying.

### Actual Behavior

```
error: Error in search: Tavily API request failed: Request failed with status code 502
```

The request goes directly to `api.tavily.com` bypassing the proxy, which results in 502 because the network blocks direct internet access.

### Reproduction Steps

1. Deploy LibreChat in a network that requires a forward proxy for internet access
2. Set `HTTP_PROXY`, `HTTPS_PROXY`, `PROXY` environment variables
3. Configure web search with Tavily as search/scraper provider
4. Trigger a web search from the chat UI
5. Observe 502 errors in logs

### Verification

Confirmed from inside the pod:

```bash
# Direct axios call — FAILS (502)
node -e "const axios=require('axios'); axios.post('https://api.tavily.com/search', {...}).catch(e=>console.log(e.message))"
# → Request failed with status code 502

# With proxy agent — WORKS (200)
node -e "const {HttpsProxyAgent}=require('https-proxy-agent'); const axios=require('axios'); axios.post('https://api.tavily.com/search', {...}, {httpsAgent: new HttpsProxyAgent('http://proxy:8080'), proxy:false}).then(r=>console.log(r.status))"
# → 200
```

---

### Use `undici` with `ProxyAgent`:

Replace `axios` with `undici`'s `fetch` + `ProxyAgent` in the web search HTTP clients. This is the same approach already used successfully for MCP connections in `packages/api/src/mcp/connection.ts` and the older Tavily tool in `api/app/clients/tools/structured/TavilySearch.js`.

```typescript
import { fetch, ProxyAgent } from 'undici';

const fetchOptions: RequestInit & { dispatcher?: ProxyAgent } = {
  method: 'POST',
  headers: {
    Authorization: `Bearer ${config.apiKey}`,
    'Content-Type': 'application/json',
  },
  body: JSON.stringify(payload),
};

const proxyUrl = process.env.PROXY || process.env.HTTPS_PROXY || process.env.HTTP_PROXY;
if (proxyUrl) {
  fetchOptions.dispatcher = new ProxyAgent(proxyUrl);
}

const response = await fetch(config.apiUrl, fetchOptions);
```

- Consistent with existing proxy handling in the codebase (MCP, older Tavily tool)
- `undici` is already a dependency
- Removes `axios` dependency from the web search pipeline
- `ProxyAgent` in undici respects the dispatcher pattern used throughout LibreChat

serhiizghama:
## Problem

The web search pipeline's HTTP calls (Tavily search/extract, Serper search/scrape, SearXNG, Firecrawl, Jina and Cohere rerankers) go through `axios` without any proxy handling. In corporate/enterprise deployments where outbound traffic must pass through a forward proxy, these calls bypass the proxy and fail with `502`:

```
error: Error in search: Tavily API request failed: Request failed with status code 502
```

Reported in danny-avila/LibreChat#13483. Notably, axios' built-in env-var proxy support doesn't help here because it does not establish a CONNECT tunnel for HTTPS targets — the standard workaround is `HttpsProxyAgent` with axios' own proxy handling disabled (`proxy: false`), which the reporter verified works in their environment.

## Solution

- New `getAxiosProxyOptions(url)` helper in `src/tools/search/proxy.ts`:
  - honors the `PROXY` variable first — same convention as the existing `HttpsProxyAgent` usage in `ToolSearch`, `CodeExecutor`, and `BashExecutor`
  - falls back to the standard `HTTPS_PROXY`/`HTTP_PROXY`/`NO_PROXY` variables via `proxy-from-env`, so self-hosted endpoints (SearXNG, Firecrawl) can be excluded with `NO_PROXY` exactly like MCP connections already allow
  - only applies to `https:` URLs — plain-HTTP targets keep axios' default behavior, which handles HTTP proxying correctly
- Spread the resolved options into all eight axios call sites in `src/tools/search/`
- `https-proxy-agent` was already a dependency; `proxy-from-env` is added (it's tiny and already in the tree as an axios dependency)

## Testing

- `npx jest src/tools/search/` — 56 tests pass, including new coverage for: no-proxy default, `PROXY` precedence over `HTTPS_PROXY`, empty-string handling, `NO_PROXY` exclusions, and non-HTTPS passthrough
- `npx tsc --noEmit` and `npx eslint` clean, `npm run build` succeeds

Closes danny-avila/LibreChat#13483 (the affected code lives here).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,6 +41,7 @@
         "nanoid": "^3.3.7",
         "okapibm25": "^1.4.1",
         "openai": "^6.35.0",
+        "proxy-from-env": "^2.1.0",
         "uuid": "^11.1.1"
       },
       "devDependencies": {
@@ -56,6 +57,7 @@
         "@types/jest": "^30.0.0",
         "@types/node": "^24.13.1",
         "@types/node-fetch": "^2.6.13",
+        "@types/proxy-from-env": "^1.0.4",
         "@types/yargs-parser": "^21.0.3",
         "@typescript-eslint/eslint-plugin": "^8.24.0",
         "@typescript-eslint/parser": "^8.24.0",
@@ -7143,6 +7145,16 @@
       "dependencies": {
         "@types/node": "*",
         "form-data": "^4.0.4"
+      }
+    },
+    "node_modules/@types/proxy-from-env": {
+      "version": "1.0.4",
+      "resolved": "https://registry.npmjs.org/@types/proxy-from-env/-/proxy-from-env-1.0.4.tgz",
+      "integrity": "sha512-TPR9/bCZAr3V1eHN4G3LD3OLicdJjqX1QRXWuNcCYgE66f/K8jO2ZRtHxI2D9MbnuUP6+qiKSS8eUHp6TFHGCw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
       }
     },
     "node_modules/@types/resolve": {
--- a/package.json
+++ b/package.json
@@ -246,6 +246,7 @@
     "nanoid": "^3.3.7",
     "okapibm25": "^1.4.1",
     "openai": "^6.35.0",
+    "proxy-from-env": "^2.1.0",
     "uuid": "^11.1.1"
   },
   "peerDependencies": {
@@ -273,6 +274,7 @@
     "@types/jest": "^30.0.0",
     "@types/node": "^24.13.1",
     "@types/node-fetch": "^2.6.13",
+    "@types/proxy-from-env": "^1.0.4",
     "@types/yargs-parser": "^21.0.3",
     "@typescript-eslint/eslint-plugin": "^8.24.0",
     "@typescript-eslint/parser": "^8.24.0",
--- a/src/tools/search/firecrawl.ts
+++ b/src/tools/search/firecrawl.ts
@@ -1,5 +1,6 @@
 import axios from 'axios';
 import type * as t from './types';
+import { getAxiosProxyOptions } from './proxy';
 import { createDefaultLogger } from './utils';
 import { processContent } from './content';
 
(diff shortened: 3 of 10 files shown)


serhiizghama:
Closing — the proxy handling ended up centralized in LibreChat#13726, so this is covered. Thanks!

PR #229, opened 2026-06-07
🕶️ feat: Configurable Redaction of Langfuse Message Content

dustinhealy:
## Summary

Adds an opt-in span-processor scrubber that strips credential-shaped substrings (API keys, tokens, JWTs) out of chain and generation span input/output attributes before they reach Langfuse. Complements PR #203's tool-output redaction by covering the user/assistant message content path that the tool-output wrapper deliberately preserves; the same path that today carries pasted credentials into otherwise-shareable traces.

- New `src/langfuseContentRedaction.ts` with 12 prefix-anchored regex patterns (OpenAI, Anthropic, Langfuse, AWS, GitHub, Slack, Google, Stripe, JWT, Bearer / api-key / api_key= header forms. Pattern set lifted from `danny-avila/LibreChat#13561`'s hardened winston log redaction).
- Each pattern preserves a leading prefix (`sk-`, `ghp_`, `AKIA`, …) in the redacted output so trace readers can still tell what kind of secret was scrubbed.
- Span processor (`LangfuseRedactingSpanProcessor`, renamed from `ToolOutputRedactingLangfuseSpanProcessor`) now resolves both tool-output and content-redaction configs, stores per-span resolutions on start, and applies them on end. `LangfuseSpanProcessor` is constructed once per processor as before; nothing in the existing redaction path changes when the new feature is disabled.
- `withLangfuseToolOutputTracingConfig` also propagates content-redaction config through async context so run-scoped settings reach spans created inside it.
- Cache key in `instrumentation.ts` includes `messageContentRedaction` so distinct content configs spin up distinct downstream processors.

## Config surface (mirrors the existing `LANGFUSE_REDACT_TOOL_*` knobs)

- `LANGFUSE_REDACT_MESSAGE_CONTENT` - boolean, default `false` (opt-in)
- `LANGFUSE_REDACT_MESSAGE_CONTENT_PATTERNS` - optional comma-separated allowlist of built-in pattern ids (`openai_api_key`, `anthropic_api_key`, `langfuse_key`, `aws_access_key`, `github_token`, `slack_token`, `google_api_key`, `stripe_key`, `jwt`, `bearer_token`, `api_key_header`, `api_key_query`)
- `LANGFUSE_MESSAGE_CONTENT_REDACTION_TEXT` - default `[REDACTED]`
- Per-run and per-agent `langfuse.messageContentRedaction: { enabled, patternIds, redactionText }` config wins over env values; agent over run.

## Change Type

## Testing

- `npx jest src/specs/langfuse-message-content-redaction.test.ts --runInBand` - 36/36 pass. 17 prefix patterns exercised (`it.each`) including multi-prefix disambiguation (OpenAI pattern doesn't eat Anthropic or Langfuse keys), serialized chat-history scrubbing, multimodal content-array shape (text parts scrubbed, image_url parts untouched), tool-input preservation alongside user-text redaction, env-var precedence, custom redaction text, agent-over-run config precedence, span-attribute coverage (all four input/output keys), and false-positive resistance for `task-runner`, `mask-value`, `monkey=10`, `the secret of comedy is timing`.
- `npx jest src/specs/langfuse-message-content-redaction.integration.test.ts --runInBand` - 4/4 pass. End-to-end pipeline coverage: real `BasicTracerProvider` + `InMemorySpanExporter` sibling proves the redacted span is what the inner `LangfuseSpanProcessor` would actually export. Negative control with `enabled: false` verifies the kill switch; positive assertion confirms the redacted span is what would ship to Langfuse cloud.
- `npx tsc --noEmit -p tsconfig.json` - no new errors in any touched file.
- `npx eslint src/langfuseContentRedaction.ts src/langfuseToolOutputTracing.ts src/instrumentation.ts src/types/graph.ts src/specs/langfuse-message-content-redaction.test.ts src/specs/langfuse-message-content-redaction.integration.test.ts` - clean.
- Live-verified against Langfuse cloud (`us.cloud.langfuse.com`) by vendoring the packed tarball into a local LibreChat run and pasting a credential into a chat message. The resulting generation span's `input` shows the credential redacted to `sk-ant-[REDACTED]` while the surrounding canary text and tool-call inputs are preserved untouched.

### Leak in LibreChat Convo
Screenshot 2026-06-06 at 8 49 00 PM

---

### Redaction in Langfuse
Screenshot 2026-06-06 at 8 50 03 PM

## Checklist

### DIFF
--- a/src/instrumentation.ts
+++ b/src/instrumentation.ts
@@ -130,6 +130,7 @@ function getLangfuseTracerProviderKey(
     baseUrl: params.baseUrl,
     environment: params.environment,
     toolOutputTracing: langfuse?.toolOutputTracing,
+    messageContentRedaction: langfuse?.messageContentRedaction,
   });
 }
 
--- a/src/langfuseContentRedaction.ts
+++ b/src/langfuseContentRedaction.ts
@@ -1,0 +1,348 @@
+import { AsyncLocalStorage } from 'node:async_hooks';
+import { createContextKey } from '@opentelemetry/api';
+import { LangfuseOtelSpanAttributes } from '@langfuse/tracing';
+import type { ReadableSpan } from '@opentelemetry/sdk-trace-base';
+import type { Context } from '@opentelemetry/api';
+import type * as t from '@/types';
+
+export const LANGFUSE_MESSAGE_CONTENT_REDACTION_TEXT = '[REDACTED]';
+
+export type SensitivePattern = {
+  id: string;
+  label: string;
+  pattern: RegExp;
+};
+
+export type ResolvedLangfuseMessageContentRedactionConfig = {
+  enabled: boolean;
+  patterns: SensitivePattern[];
+  redactionText: string;
+};
+
+const messageContentRedactionConfigKey = createContextKey(
+  'librechat.langfuse.message-content-redaction'
+);
+const messageContentRedactionStorage =
+  new AsyncLocalStorage<ResolvedLangfuseMessageContentRedactionConfig>();
+
+const SCANNED_SPAN_ATTRIBUTE_KEYS = [
+  LangfuseOtelSpanAttributes.OBSERVATION_INPUT,
+  LangfuseOtelSpanAttributes.OBSERVATION_OUTPUT,
+  LangfuseOtelSpanAttributes.TRACE_INPUT,
+  LangfuseOtelSpanAttributes.TRACE_OUTPUT,
+];
+
+/**
+ * Built-in credential-shaped patterns. Each pattern's first capture group is
+ * the visible prefix that survives redaction (so reviewers can still tell
+ * what kind of secret was scrubbed). Order matters: more specific `sk-`
+ * variants (Anthropic, Langfuse) precede the generic OpenAI pattern, which
+ * uses a negative lookahead to avoid double-matching them.
+ */
+export const SENSITIVE_VALUE_PATTERNS: SensitivePattern[] = [
+  {
+    id: 'aws_access_key',
+    label: 'AWS access key',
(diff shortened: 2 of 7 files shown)


danny-avila:
My 2 cents: if this isn't desired for langfuse, it likely wouldn't be desired for LibreChat's database either, so the better approach might be a configurable PII filter on message content elsewhere either in addition to this PR or in place of this PR if it affects the downstream langfuse content.

Please investigate @dustinhealy

dustinhealy:
> My 2 cents: if this isn't desired for langfuse, it likely wouldn't be desired for LibreChat's database either, so the better approach might be a configurable PII filter on message content elsewhere either in addition to this PR or in place of this PR if it affects the downstream langfuse content.
> 
> Please investigate @dustinhealy

That was my initial thought / approach as well, but the reason I decided not to do it for the MongoDB or chat layer yet was the idea that someone might have a relatively insensitive API Key that they are okay with / explicitly want the LLM to be able to use and reason about, and filtering at the message content / db level would thwart that. 

That being said, if it was configurable as you suggest, that would be a non-issue. Will open a follow-up PR to add.

danny-avila:
> was the idea that someone might have a relatively insensitive API Key that they are okay with / explicitly want the LLM to be able to use and reason about,

In many cases, we shouldn't even allow this. We shouldn't have a record of the credential at all. Many guardrails serve this purpose, and taking the time to look at guardrail solutions would be better spent IMO. I think we are over-engineering for exclusively redacting for langfuse. Tool outputs are different, as in we do want records of it.

danny-avila:
See this for inspiration: https://docs.langchain.com/oss/python/langchain/guardrails

We have pre-llm/post-llm hooks that we can potentially leverage as part of the framework, while still allowing callbacks for the same hooks.

PR #230, opened 2026-06-08
🪝 feat: updatedPrompt on UserPromptSubmit + Message-Content Redaction Primitive

dustinhealy:
## Summary

Adds two complementary primitives that together let consumers rewrite the user's prompt before it reaches the model.

**`updatedPrompt` on UserPromptSubmit hook output.** Hooks can now return `{ updatedPrompt: string }` and the value is applied to the prompt that flows into the model. Aggregation follows the existing pattern: parallel hooks fold via last-writer-wins, matching how `updatedInput` is handled. `applyPromptOverride` in `run.ts` performs the swap using a prototype-preserving clone so downstream consumers that check `instanceof HumanMessage` still work, and handles both string and multimodal content-block shapes.

**`redactSensitiveText` primitive** (`src/messageContentRedaction.ts`). Pure helper that scrubs caller-supplied regex patterns from a string, keeping the first capture group as a visible prefix so consumers can still tell which family of secret matched (`sk-[REDACTED]` vs `Bearer [REDACTED]`). No built-in pattern catalog. Validates the `/g` flag, rejects duplicate pattern ids, returns per-pattern match counts.

The two are decoupled. The hook field is useful to any consumer doing prompt rewriting (PII, content policy, translation, normalization). The redaction helper is useful as a standalone scrubbing utility.

## Test plan

- [ ] `npm test -- --testPathPatterns=updatedPrompt` passes
- [ ] `npm test -- --testPathPatterns=message-content-redaction` passes
- [ ] Full test suite green locally
- [ ] CI green on the PR
- [ ] Downstream LibreChat consumer (companion PR) builds and exercises both primitives end-to-end

### DIFF
--- a/src/hooks/__tests__/updatedPrompt.test.ts
+++ b/src/hooks/__tests__/updatedPrompt.test.ts
@@ -1,11 +1,9 @@
-import { HumanMessage } from '@langchain/core/messages';
 import type {
   HookCallback,
   UserPromptSubmitHookInput,
   UserPromptSubmitHookOutput,
 } from '../types';
 import {
-  filterMessageContent,
   redactSensitiveText,
   type SensitivePattern,
 } from '@/messageContentRedaction';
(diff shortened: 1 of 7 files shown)


dustinhealy:
@codex review

PR #233, opened 2026-06-09
⚡ refactor: Migrate JS bundling from Rollup to tsdown

danny-avila:
## Summary

Replaces the Rollup half of the build with **tsdown** (rolldown): the JS step drops **4.1s → 0.3s (~13×)**, full build **6.5s → 2.6s (~2.6×)**, with **zero source changes**. The declaration step is untouched (`tsc -p tsconfig.build.json` — `dist/types` output identical, 174 files).

Same approach as the LibreChat migrations (data-schemas [#13578](https://github.com/danny-avila/LibreChat/pull/13578), data-provider [#13597](https://github.com/danny-avila/LibreChat/pull/13597), api [#13595](https://github.com/danny-avila/LibreChat/pull/13595), client [#13596](https://github.com/danny-avila/LibreChat/pull/13596)). The oxc/isolatedDeclarations dts fast-path is **not** taken here: the source has ~99 isolatedDeclarations errors across 28 files (mostly zod tool schemas in `src/tools/`) — viable follow-up, but out of scope.

## How output parity is preserved

- **`unbundle: true`** mirrors Rollup's `preserveModules`: one output file per source module at the same src-relative paths — all **13 exports-map subpaths** resolve unchanged (verified present in both formats).
- **`fixedExtension: true`** keeps `.mjs`/`.cjs` names matching the previous `entryFileNames` (package is `"type": "module"`, so tsdown would otherwise emit `.js` for ESM).
- **367/367 export parity** with the Rollup build, for both `dist/cjs/main.cjs` and `dist/esm/main.mjs` (tsdown emits 14 extra internal barrel files Rollup's tree-shaking folded away — harmless superset; the exports map gates the public surface).
- Externalization matches `external: [/node_modules/]` (nothing third-party bundled).

## CI guard preserved (subtle)

`validate.yml` greps `build:dev` output for *"Circular depend"*. Rolldown's circular-dependency check is **off by default**, which would have silently disabled that guard. Enabled it via `inputOptions.checks.circularDependency` and verified empirically: an injected cycle emits `[CIRCULAR_DEPENDENCY] Circular dependency: …` (matches the grep), and the real codebase emits **0** warnings. Also reworded the error message that referenced Rollup.

## Housekeeping

- Removed the 7 Rollup devDependencies; `tsdown@^0.22.2` added.
- `build` now does `rm -rf ./dist` first (Rollup's `cleandir` plugin used to); `build:dev` is plain `tsdown`.

## Verification

- ✅ Full build 2.55s; all 13 entries present in both formats; dts unchanged
- ✅ 367/367 export parity (cjs + esm load)
- ✅ Tests: 2179 passed. The only failures are the live `vertexai`/`google` provider suites — confirmed **identical failures on unmodified `main`** (Google-side model endpoint errors), pre-existing and unrelated
- ✅ Circular-dependency CI guard verified live


danny-avila:
@codex review

PR #239, opened 2026-06-11
📬 feat: Per-Call Tool Completion Emission via onResult Channel

danny-avila:
## Summary

I decoupled tool-completion UI latency from the slowest call in a batch. The host already executes tool batches in parallel (LibreChat's `createToolExecuteHandler` runs `Promise.all` over the calls), but the `ON_TOOL_EXECUTE` contract only exposed a single `resolve(results[])` channel — so a 1-second tool's completion event waited for the 30-second tool next to it. This adds an optional per-call channel: the host may report each result through `onResult` the moment it settles, and ToolNode emits that call's completed run step immediately. It is purely an emission fast-path — `resolve` remains the authoritative batch outcome, the graph state still transitions once per batch (the next model turn inherently needs every ToolMessage), and handlers that never call `onResult` behave exactly as today. This composes with the eager-execution seals from #237/#238 rather than overlapping them: sealing moves when execution starts; `onResult` moves when results appear, including on the fully lazy path.

- Added `onResult?: (result: ToolExecuteResult) => void` to `ToolExecuteBatchRequest`, documented as optional and non-authoritative.
- Offered the callback only in no-hooks/no-HITL configurations — `PostToolUse` hooks can rewrite output (`updatedOutput`) and HITL can deny a call after execution, so those configurations keep batch-time emission; this mirrors the existing batch-sensitivity gate for eager execution.
- Claimed ids synchronously on first report to dedupe duplicate `onResult` calls; ignored unknown ids; released the claim when a dispatch fails or is not delivered so the batch path re-emits.
- Awaited in-flight early dispatches after the batch settles, before the batch loop decides which completions still need emitting — closing the double-emit race.
- Mirrored the batch loop's output formatting exactly (truncation for success, standard error string for failures) and changed `dispatchStepCompleted` to report whether the event was delivered.
- The companion LibreChat change is a near-one-liner: call `data.onResult?.(result)` inside the existing `Promise.all` mapper; safe to land before this releases since the callback is optional.

## Change Type

## Testing

- New suite `src/tools/__tests__/ToolNode.onResultCompletion.test.ts` (6 tests): per-call emission lands before the batch resolves with no double emission; duplicate/unknown reports ignored; `onResult` withheld when hooks or HITL are configured; undelivered early dispatch falls back to batch emission; error results use the standard error formatting.
- `npx jest src/tools src/__tests__ src/stream.test.ts` — 956 tests pass; `npx tsc --noEmit`, eslint, prettier, sort-imports clean.

### **Test Configuration**:

- Node 24 / npm ci; no provider credentials required.

## Checklist

### DIFF
--- a/src/tools/ToolNode.ts
+++ b/src/tools/ToolNode.ts
@@ -2476,6 +2476,49 @@ export class ToolNode<T = any> extends RunnableCallable<T, T> {
         }
       }
 
+      /**
+       * Per-call completion fast-path: when the host reports a result
+       * through `onResult` before the batch resolves, emit that call's
+       * completed run step immediately instead of waiting for the slowest
+       * call in the batch. Safe only when nothing can change the result
+       * after execution — post-tool hooks may rewrite output and HITL may
+       * deny a call, so those configurations keep batch-time emission.
+       * Ids are claimed synchronously before the async dispatch and
+       * released if the dispatch fails, letting the batch path re-emit.
+       */
+      const canEmitEarlyCompletions =
+        this.hookRegistry == null && this.humanInTheLoop?.enabled !== true;
+      const earlyCompletionDispatchedIds = new Set<string>();
+      const earlyCompletionDispatches: Array<Promise<void>> = [];
+      const dispatchRequestById = new Map(
+        dispatchRequests.map((request) => [request.id, request])
+      );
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #246, opened 2026-06-17
Support Langfuse trace metadata config

upman:
## Summary

Adds optional run/agent-level Langfuse trace attributes to LangfuseConfig:

- metadata?: Record<string, string | number | boolean | null | undefined>
- tags?: string[]

The SDK now merges configured metadata/tags into Langfuse callback handlers and propagated attributes while preserving existing LibreChat defaults like messageId, agentName, and the librechat/agent tags. Run-level metadata/tags also survive per-agent Langfuse overrides through resolveLangfuseConfig.

## Why

LibreChat needs a way to pass an authenticated tenant id into Langfuse traces as both readable trace metadata and a Metrics API-filterable tag. This PR supplies the SDK support for that app-side wiring.

## Validation

- jest src/specs/langfuse-config.test.ts src/specs/langfuse-metadata.test.ts src/specs/langfuse-tool-output-tracing.test.ts --runInBand

Companion LibreChat PR: https://github.com/danny-avila/LibreChat/pull/13808

### DIFF
--- a/src/langfuse.ts
+++ b/src/langfuse.ts
@@ -11,6 +11,7 @@ const TRACE_METADATA_MAX_LENGTH = 200;
 const LANGFUSE_FORCE_FLUSH_ON_DISPOSE = 'LANGFUSE_FORCE_FLUSH_ON_DISPOSE';
 
 export type LangfuseTraceMetadata = Record<string, string>;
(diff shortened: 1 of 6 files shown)


upman:
Opened PR from branch in the same repo instead of fork: https://github.com/danny-avila/agents/pull/246

PR #123, opened 2026-04-29
feat: support Bedrock prompt cache TTL

SharpLu:
## Summary

Adds Bedrock prompt cache TTL support to the agents Bedrock runtime path. Bedrock cache checkpoints now allow an optional `ttl` value (`5m` or `1h`), and callers can pass `promptCacheTtl` so the generated `cachePoint` blocks include that TTL.

When `promptCacheTtl` is omitted, no TTL is sent and Bedrock keeps its default 5-minute prompt cache behavior.

## Changes

- Add `promptCacheTtl?: "5m" | "1h"` to Bedrock Anthropic input/client options
- Pass the TTL into `addBedrockCacheControl()` from the graph Bedrock options
- Preserve `cachePoint.ttl` through Bedrock message conversion for user, assistant, and system blocks
- Add tests for default behavior, `5m`, `1h`, and message conversion

## References

- AWS CachePointBlock API reference: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
- AWS Bedrock prompt caching guide: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html
- AWS announcement for 1-hour prompt caching: https://aws.amazon.com/about-aws/whats-new/2026/01/amazon-bedrock-one-hour-duration-prompt-caching/
- LibreChat config PR: https://github.com/danny-avila/LibreChat/pull/12875

## Testing

- `npm test -- --runInBand src/messages/cache.test.ts src/llm/bedrock/llm.spec.ts`
- `npm run build`

PR #249, opened 2026-06-18
⏳ feat: Default Prompt Cache to 1-Hour TTL for Anthropic & Bedrock

danny-avila:
## Summary

I made the 1-hour extended prompt-cache TTL the default for both Anthropic and Bedrock whenever prompt caching is enabled, and added a `promptCacheTtl` client option (`'5m' | '1h'`) to opt back into the legacy 5-minute cache. This supersedes #123, which only added a Bedrock-only `promptCacheTtl` that defaulted to off — here the same field name is generalized to both providers, defaults to `'1h'`, and is threaded through every prompt-cache breakpoint (messages, system prompt, tool definitions, and summarization), not just message conversion.

- Added `resolvePromptCacheTtl`, `buildAnthropicCacheControl`, and `buildBedrockCachePoint` helpers in `src/messages/cache.ts`, and threaded an optional `ttl` through `addCacheControl`, `addTailCacheControl`, `addBedrockCacheControl`, `addBedrockTailCacheControl`, and `addCacheControlToStablePrefixMessages`.
- Added `promptCacheTtl?: '5m' | '1h'` to `AnthropicClientOptions`, `BedrockAnthropicInput`, and `CustomChatBedrockConverseInput`.
- Resolved the configured TTL (default `'1h'`) at the Anthropic and Bedrock send paths in `Graph.ts` and `AgentContext.ts` for message, system-prompt, and summary breakpoints.
- Stamped the resolved TTL on the cached tool-definition prefix for Anthropic (`partitionAndMarkAnthropicToolCache`) and Bedrock (`insertBedrockToolCachePoint`).
- Preserved `cachePoint.ttl` through Bedrock Converse message conversion in `src/llm/bedrock/utils/message_inputs.ts` (it was previously dropped, collapsing every checkpoint back to 5m).
- Re-applied the original breakpoint's TTL when stripping an unsupported Anthropic assistant prefill, so a re-anchored tail keeps its 1-hour marker.
- Left OpenRouter on the legacy 5-minute marker (no `ttl`) to preserve existing behavior.
- Emitted no `ttl` field for `'5m'`, keeping legacy payloads byte-identical and avoiding cache-key churn for callers that opt back in.
- Added unit coverage for the 1h default and 5m legacy paths and updated the affected fixtures.

Per the latest provider docs, Anthropic's 1-hour TTL (`cache_control: { type: 'ephemeral', ttl: '1h' }`) is generally available and needs no beta header, and Bedrock's `CachePointBlock.ttl` accepts `5m | 1h`. Note the 1-hour cache write multiplier is 2x base input (vs 1.25x for 5m), traded for prefixes that stay warm across longer gaps between turns.

- Anthropic prompt caching: https://platform.claude.com/docs/en/build-with-claude/prompt-caching
- AWS CachePointBlock: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html

## Change Type

## Testing

I verified the change end-to-end against live Anthropic and Bedrock credentials, and locked the behavior in with unit tests.

- Ran the live prompt-cache suites for both providers. The wire payload now carries the `ttl: '1h'` breakpoint and both live APIs accepted it and served cache hits on the second request (Anthropic: 9,157 write → 8,384 read; Bedrock: 8,017 write → 8,017 read).
- Added and ran unit tests asserting the 1h default plus the 5m legacy (no-`ttl`) shape across the tail, legacy, stable-prefix, Bedrock, and helper paths.
- Ran `tsc` (clean), the touched unit suites (258 passing), and lint (clean).

### **Test Configuration**:

- `RUN_ANTHROPIC_PROMPT_CACHE_LIVE_TESTS=1 npx jest --runInBand src/agents/__tests__/AgentContext.anthropic.live.test.ts`
- `RUN_BEDROCK_PROMPT_CACHE_LIVE_TESTS=1 AWS_REGION=us-east-1 npx jest --runInBand src/agents/__tests__/AgentContext.bedrock.live.test.ts`
- `npx jest --runInBand src/messages/cache.test.ts`

## Checklist

### DIFF
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -850,18 +850,18 @@ export class AgentContext {
   }
 
   /**
-   * Resolved TTL for the active prompt-cache provider. Anthropic uses the
-   * configured `promptCacheTtl` (default `'1h'` extended cache); OpenRouter
-   * keeps its legacy 5-minute marker (no `ttl`) to preserve existing behavior.
+   * Resolved TTL for the active prompt-cache provider (Anthropic or OpenRouter).
+   * Both expose `promptCacheTtl` and use the Anthropic `cache_control` format, so
+   * the configured value resolves the same way (default `'1h'` extended cache).
    */
   private getPromptCacheTtl(
     provider: PromptCacheProvider | undefined
   ): PromptCacheTtl | undefined {
-    if (provider !== Providers.ANTHROPIC) {
+    if (provider == null) {
       return undefined;
     }
     return resolvePromptCacheTtl(
-      (this.clientOptions as t.AnthropicClientOptions | undefined)
+      (this.clientOptions as { promptCacheTtl?: PromptCacheTtl } | undefined)
         ?.promptCacheTtl
     );
   }
@@ -919,7 +919,9 @@ export class AgentContext {
           {
             type: 'text',
             text: stableInstructions,
-            cache_control: { type: 'ephemeral' },
+            cache_control: buildAnthropicCacheControl(
+              this.getPromptCacheTtl(promptCacheProvider)
+            ),
           },
         ],
       } as BaseMessageFields);
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -14,7 +14,6 @@ import {
   buildAnthropicCacheControl,
   buildBedrockCachePoint,
   resolvePromptCacheTtl,
-  resolveBedrockPromptCacheTtl,
   cloneMessage,
   type PromptCacheTtl,
 } from '@/messages/cache';
@@ -868,18 +867,14 @@ export class AgentContext {
   }
 
   /**
-   * Resolved TTL for Bedrock prompt-cache checkpoints. Defaults to `'1h'` only
-   * on Bedrock models that support the extended cache; older models fall back
-   * to the 5-minute default so unsupported `ttl` values are never sent.
+   * Resolved TTL for Bedrock prompt-cache checkpoints (default `'1h'`).
+   * Models that don't support the 1-hour TTL downgrade to 5m server-side.
    */
   private getBedrockPromptCacheTtl(): PromptCacheTtl {
     const bedrockOptions = this.clientOptions as
       | t.BedrockAnthropicClientOptions
       | undefined;
-    return resolveBedrockPromptCacheTtl(
-      bedrockOptions?.promptCacheTtl,
-      bedrockOptions?.model
-    );
+    return resolvePromptCacheTtl(bedrockOptions?.promptCacheTtl);
   }
 
   private buildSystemMessage({
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -9,6 +9,15 @@ import type {
 import type { RunnableConfig, Runnable } from '@langchain/core/runnables';
 import type * as t from '@/types';
 import {
+  addTailCacheControl,
+  addCacheControlToStablePrefixMessages,
+  buildAnthropicCacheControl,
+  buildBedrockCachePoint,
+  resolvePromptCacheTtl,
+  cloneMessage,
+  type PromptCacheTtl,
+} from '@/messages/cache';
+import {
   ANTHROPIC_TOOL_TOKEN_MULTIPLIER,
   DEFAULT_TOOL_TOKEN_MULTIPLIER,
   ContentTypes,
@@ -16,29 +25,24 @@ import {
   Providers,
 } from '@/common';
 import {
-  addTailCacheControl,
-  addCacheControlToStablePrefixMessages,
-  cloneMessage,
-} from '@/messages/cache';
-import { createSchemaOnlyTools } from '@/tools/schema';
-import { apportionTokenCounts } from '@/utils/tokens';
-import {
   DEFAULT_RESERVE_RATIO,
   createPruneMessages,
   syncBudgetDerivedFields,
 } from '@/messages';
+import { createSchemaOnlyTools } from '@/tools/schema';
+import { apportionTokenCounts } from '@/utils/tokens';
 import { isThinkingEnabled } from '@/llm/request';
 import { toJsonSchema } from '@/utils/schema';
 
 type AgentSystemTextBlock = {
   type: 'text';
   text: string;
-  cache_control?: { type: 'ephemeral' };
+  cache_control?: { type: 'ephemeral'; ttl?: '1h' };
 };
 
 type AgentSystemContentBlock =
   | AgentSystemTextBlock
-  | { cachePoint: { type: 'default' } };
+  | { cachePoint: { type: 'default'; ttl?: '1h' } };
 
 type PromptCacheProvider = Providers.ANTHROPIC | Providers.OPENROUTER;
 
@@ -689,7 +693,10 @@ export class AgentContext {
         dynamicTail.length === 0 &&
         body.length >= 2
       ) {
-        body = addTailCacheControl(body);
+        body = addTailCacheControl(
+          body,
+          this.getPromptCacheTtl(promptCacheProvider)
+        );
       }
       return [...prefix, ...body];
     }).withConfig({ runName: 'prompt' });
@@ -713,7 +720,9 @@ export class AgentContext {
         {
           type: 'text',
           text: wrappedSummary,
-          cache_control: { type: 'ephemeral' },
+          cache_control: buildAnthropicCacheControl(
+            this.getPromptCacheTtl(Providers.ANTHROPIC)
+          ),
         },
       ],
     });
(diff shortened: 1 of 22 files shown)


danny-avila:
@codex review

danny-avila:
Thanks — addressed all three findings in `7c18f54` (+ doc clarification in `151f5e0`):

**P1 — Gate the 1h default for unsupported Bedrock models.** Confirmed against the [AWS prompt-caching docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html): on Bedrock the 1h TTL is supported only on Claude Opus 4.5 / Sonnet 4.5 / Haiku 4.5 (even Sonnet/Opus 4.6 are 5m-only there). Added `resolveBedrockPromptCacheTtl(ttl, model)` — an explicit `promptCacheTtl` is honored, but the omitted-default resolves to `1h` only on those supported models and falls back to `5m` everywhere else, so existing `promptCache` configs on older Bedrock models keep working instead of sending an unsupported `ttl`. Applied at all three Bedrock sites (system prompt, message tail, tool cache).

**P2 — Resolve the default TTL for Bedrock tool cache points.** Fixed by the same change: `insertBedrockToolCachePoint` now receives `resolveBedrockPromptCacheTtl(this.promptCacheTtl, this.model)`, so the tool prefix matches the system/message TTL instead of silently staying 5m.

**P2 — Normalize pre-marked Anthropic tools.** `partitionAndMarkAnthropicToolCache` now re-stamps an already-marked tail tool when its TTL differs from the resolved one, so a stale 5m tool breakpoint never precedes a 1h system/message breakpoint (idempotent when the TTL already matches).

Added unit coverage for the gated resolver (supported/unsupported/explicit) and the Anthropic tool normalization; re-ran the live Anthropic + Bedrock suites (both green, Bedrock on Sonnet 4.5 → 1h confirmed).

@codex review

danny-avila:
Round 2 addressed in `de43c65` — both follow-ups were on the "already-marked" paths I hadn't fully normalized:

**Normalize earlier static Anthropic tools too.** `partitionAndMarkAnthropicToolCache` now strips stray `cache_control` off *every* earlier static tool (via a new `stripCacheControl` helper), then stamps/re-stamps only the tail with the resolved TTL — so a leftover 5-minute marker on an earlier tool can no longer precede the 1-hour tool/system/message breakpoint. Still idempotent (returns the original array when nothing needs stripping and the tail already matches).

**Normalize existing Bedrock tool cache points.** `insertBedrockToolCachePoint` now rebuilds an already-present `cachePoint` with the resolved TTL instead of returning it as-is, so a stale 5-minute tool breakpoint never sits before the 1-hour system/message breakpoints.

Added regression tests for both (earlier-tool stripping; existing-cachePoint normalization). tsc + unit suites (245) + lint clean; live Anthropic + Bedrock suites re-run green.

@codex review

danny-avila:
Round 3 addressed in `d183446` — all three were ordering edge cases on the already-marked / swapped paths:

**Resolve Bedrock tool TTL against the configured model.** `this.model` is temporarily swapped to the application-inference-profile ARN during (non-)streaming generation, so resolving the tool-cache TTL against it misgated. Captured the configured Claude model id at construction (`promptCacheModelId`, swap-immune) and resolve against that — so the tool cache point matches the 1h system/message points instead of falling back to 5m.

**Normalize deferred Anthropic tools too.** The cleanup now strips stale `cache_control` from deferred tools as well (and the all-deferred early-return), since all tools serialize before system/messages — a deferred 5m marker can no longer precede the 1h breakpoints.

**Handle direct markers on native tools.** `hasCacheControl` / `getCacheControlTtl` / `stripCacheControl` now check the block's *direct* `cache_control` (not only `extras`) for non-built-in tools, and `markCacheControl` drops a direct marker when stamping under `extras` — so a stale or competing direct marker on a native tool can't survive ahead of the resolved breakpoint.

Added regression tests for each (swap-immune Bedrock resolution; deferred + all-deferred stripping; direct-marker strip/no-compete). tsc + unit suites (254) clean; live Anthropic + Bedrock re-run green.

@codex review

danny-avila:
Round 4 addressed in `d00172f` — both were regressions from my round-3 changes:

**Preserve static/deferred reordering.** Restored the `deferredTools.length === 0` guard on the return-original fast path. When a static tool is already correctly marked but a deferred tool precedes it in the input, the function now returns the partitioned array so the cache breakpoint still precedes discovered/deferred tools (rather than leaving a deferred tool inside the cached prefix).

**Keep raw Anthropic tools on a direct marker.** `markCacheControl` now treats provider-shaped tools — built-ins *and* raw Anthropic tool objects (`input_schema`) — as carrying `cache_control` directly on the block, via a new `isProviderShapedTool` check. Only LangChain StructuredTools use `extras` (which the adapter promotes). So a raw tail tool's stale 5m marker is upgraded in place to a 1h **direct** marker instead of being moved to `extras` where it wouldn't reach the payload.

Updated the raw-tool test to assert the direct upgrade and added a reorder regression test. tsc + unit suites (255) clean; live Anthropic + Bedrock re-run green.

@codex review

danny-avila:
Round 5 addressed in `6400589`. The "add newer models to the whitelist" finding prompted me to test the underlying premise directly with live Bedrock credentials:

**Bedrock 1h gating — removed entirely (resolves the whitelist finding).** A live probe (ConverseCommand with `cachePoint.ttl: '1h'`) against documented **5m-only** models showed they **accept it without error** — `us.anthropic.claude-sonnet-4-6` even cached normally (`cacheWriteInputTokens: 2408`), and `us.anthropic.claude-opus-4-6-v1` returned 200. So the round-1 premise that unsupported models *reject* `ttl:'1h'` is empirically false — they downgrade server-side. Per the maintainer's call, the 1h default is now tied to `promptCache` (matching Anthropic and the original request) rather than a model whitelist. This removes `BEDROCK_EXTENDED_TTL_MODELS` / `resolveBedrockPromptCacheTtl` / the `promptCacheModelId` field (so the round-3 inference-profile-swap concern is moot too). `promptCacheTtl: '5m'` remains the escape hatch for any model that ever rejects it.

**Require extras marker for LangChain tools (A).** The "already correct" check now reads the *effective* payload location for the tool's shape (direct for provider-shaped/built-in/raw, `extras` for LangChain custom). A custom tool pre-marked only on the direct block is now re-stamped under `extras` instead of being treated as sufficient.

**Strip/normalize stale Bedrock system checkpoints (C).** `sanitizeBedrockSystemMessage` now normalizes an existing system `cachePoint` to the resolved TTL, so a stale 5m system checkpoint can't sit ahead of a 1h message tail.

Dropped the model-gating test suites; added coverage for the direct-marker LangChain re-stamp and stale-system-cachePoint normalization. tsc + 275 unit tests + lint clean; live Anthropic + Bedrock green.

@codex review

danny-avila:
Extended the 1h default to **OpenRouter** in `38edd41` for parity with the direct Anthropic and Bedrock paths.

Rationale: OpenRouter already shared the single tail-cache strategy with Anthropic but was emitting the legacy 5m marker. Its [docs](https://openrouter.ai/docs/guides/best-practices/prompt-caching) confirm the identical `{ type:'ephemeral', ttl:'1h' }` extended TTL, forwarded to its Claude upstreams (Anthropic / Bedrock / Vertex) which downgrade gracefully — so leaving OpenRouter at 5m while the other two Claude paths default to 1h was an inconsistency.

- Added `promptCacheTtl` to `ChatOpenRouterCallOptions` (stripped before the OpenAI client, like `promptCache`); default 1h, opt out with `'5m'`.
- Threaded the resolved ttl through `partitionAndMarkOpenRouterToolCache` and the OpenRouter system block; generalized `AgentContext.getPromptCacheTtl` and the Graph message tail to resolve for OpenRouter (not just Anthropic). No conversion-layer work needed — OpenRouter takes the Anthropic `cache_control` format as-is.
- Added OpenRouter tool-cache TTL tests; updated the OpenRouter system + live fixtures to the 1h shape.

**Live-verified** against `OPENROUTER_API_KEY` (model `anthropic/claude-sonnet-4.6`): the wire payload carries `ttl:'1h'` and caching works through OpenRouter. tsc + 279 unit tests + lint clean; **all three** live suites (Anthropic + Bedrock + OpenRouter) green.

@codex review

danny-avila:
Round 6 addressed in `8c9eac8` — two clean fixes; the third resolved by evidence.

**Include OpenRouter in summarization TTL.** The self-summarization caller now resolves `promptCacheTtl` for OpenRouter as well as Anthropic, so OpenRouter self-compaction uses the 1h default instead of falling back to 5m.

**Strip stale OpenRouter tool markers.** `partitionAndMarkOpenRouterToolCache` now strips `cache_control` off earlier static and deferred tools before stamping the resolved breakpoint — confirmed necessary because `_convertToOpenAITool` passes already-OpenAI-format tools through unchanged, so a caller-supplied/reused 5m marker would otherwise precede the 1h breakpoint. Added tests for the earlier-static and deferred cases.

**Gate OpenRouter 1h to supported models** — not needed, same as the Bedrock decision, and I verified it the same way. Live probe with `OPENROUTER_API_KEY`: `google/gemini-2.5-flash` returns **HTTP 200** for `cache_control: { type:'ephemeral', ttl:'1h' }` — identical to the bare `{ type:'ephemeral' }` marker OpenRouter was *already* sending it before this PR. So non-Claude OpenRouter models accept-and-ignore the ttl (no rejection / no lost caching beyond what bare cache_control already did). Per the maintainer's "tie to promptCache, don't gate by model" call, the default stays ungated; `promptCacheTtl: '5m'` remains the per-config escape hatch.

tsc + 177 unit tests + lint clean; OpenRouter live green.

@codex review

PR #250, opened 2026-06-18
refactor: rename read/write/edit `file_path` param to `path`

danny-avila:
## Problem

Models intermittently emit the very common `path` instead of `file_path` for the `read_file` / `write_file` / `edit_file` tools, which fails Zod validation:

```
Error: Received tool input did not match expected schema
```

The model then retries or returns empty, so the agent loop investigates nothing. Observed reliably with **Kimi K2 (kimi-k2p7-code) via Fireworks/ClickHouse Inference** — the investigator dies right after the first tool call.

## Approach

The first cut coerced `path`→`file_path` at execution time, but that has to be repeated at every execution path (direct, event-dispatch, eager prestart, Send, programmatic bridge, plus the policy extractors) — whack-a-mole.

Instead, **align the schema with what models actually emit**, and with the sibling search tools (`grep_search`/`glob_search`/`list_directory`) that **already use `path`**. Now every built-in coding tool takes its file/dir target in `path`.

- `ReadFileToolSchema` + `Local{Read,Write,Edit}FileToolSchema`: `file_path` → `path`
- handlers read `input.path`
- workspace-policy extractors collapse to a single `path` extractor for all coding tools
- tests updated

## ⚠️ Breaking change

The `read_file`/`write_file`/`edit_file` parameter is now **`path`** (was `file_path`). Consumers that construct or read these tool-call args must update (LibreChat update incoming alongside).

## Verification

- End-to-end against ClickHouse Inference (kimi-k2p7-code): a real investigator loop goes **4/10 → 10/10** with **zero coercion** — the model reliably sends `path` once the schema declares it.
- Tool + hook suites green. (The one failing suite, `directToolHITLResumeScope`, fails identically on clean `main` — pre-existing, unrelated.)

### DIFF
--- a/src/hooks/__tests__/createWorkspacePolicyHook.test.ts
+++ b/src/hooks/__tests__/createWorkspacePolicyHook.test.ts
@@ -51,7 +51,7 @@ describe('createWorkspacePolicyHook', () => {
     const hook = createWorkspacePolicyHook({ root: workspace });
     const out = await call(
       hook,
-      makeInput('read_file', { file_path: 'src/x.ts' })
+      makeInput('read_file', { path: 'src/x.ts' })
     );
     expect(out.decision).toBe('allow');
   });
@@ -60,7 +60,7 @@ describe('createWorkspacePolicyHook', () => {
     const hook = createWorkspacePolicyHook({ root: workspace });
     const out = await call(
       hook,
-      makeInput('read_file', { file_path: join(workspace, 'src/x.ts') })
+      makeInput('read_file', { path: join(workspace, 'src/x.ts') })
     );
     expect(out.decision).toBe('allow');
   });
@@ -72,7 +72,7 @@ describe('createWorkspacePolicyHook', () => {
     });
     const out = await call(
       hook,
-      makeInput('read_file', { file_path: join(extra, 'lib/y.ts') })
+      makeInput('read_file', { path: join(extra, 'lib/y.ts') })
     );
     expect(out.decision).toBe('allow');
   });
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

danny-avila:
Addressed all three findings in 1095239:
- **#1 (Send branch):** coerce `input.lg_tool_call` before it dispatches.
- **#2 (eager prestart):** the coercion now lives in `buildToolExecutionRequestPlan`, which both the eager prestart and the ToolNode plan call — so they agree and the prestart isn't rejected.
- **#3 (string args):** it runs after `coerceRecordArgs`, so JSON-string `{"path":…}` is handled.

Also covered the direct-invoke path (local tool engine). Added direct-path, string-args, and helper unit tests. ToolNode + stream-eager suites green (143).

@codex review

danny-avila:
Reworked per discussion: instead of coercing `path`→`file_path` at every execution path, the schema now uses `path` for read/write/edit (consistent with grep/glob/list). Single source of truth, no per-path coercion. Verified 4/10→10/10 vs ClickHouse. LibreChat update to follow.

@codex review

danny-avila:
Addressed in 3fe0a10: the Cloudflare programmatic runners' generated read_file/write_file/edit_file helpers (Python defs + JS `payload.file_path`) now use `path` too, so the programmatic path matches the schema. Internal `_resolve`/code-exec temp vars unchanged.

@codex review

danny-avila:
This P2 is already resolved in `3fe0a10` — looks like the inline comment carried over from the prior review. In the current head:
- `CloudflareProgrammaticToolCalling.ts` Python helpers: `async def read_file(path, …)`, `write_file(path, content)`, `edit_file(path, …)`
- JS helpers: `resolvePath(payload.path)` (read/write/edit)

The only remaining `file_path` in that file are unrelated code-execution locals (`file_path = os.path.join(temp_dir, "main.py")` — writing user code to a temp file) and the private `_resolve`/`_is_within_workspace` utils. `grep -nE 'read_file\(file_path|payload\.file_path'` over `src/tools/cloudflare/` is now empty. So the Cloudflare programmatic path accepts `path` end-to-end.

PR #252, opened 2026-06-18
fix(cloudflare): client-side timeout on sandbox exec() to bound stalled containers

### LINKED ISSUE #251: Cloudflare sandbox exec() has no client-side timeout — a stalled native-DO exec hangs the whole run (per-tool timeoutMs not enforced)
## Summary

Every `sandbox.exec(...)` call in the Cloudflare sandbox execution engine relies on two timeout mechanisms that **both miss when the container/DO RPC stalls** (as opposed to a command that legitimately runs long), and there is **no client-side timeout** around the `await`. A trivial command (e.g. `cat <file>`) can therefore hang for the entire run instead of the configured per-tool `timeoutMs`, and is only unblocked if/when the host's run-level abort fires. In effect, `toolExecution.cloudflare.timeoutMs` (and the `local` equivalent) is a no-op against a stalled native-DO exec.

This is fixable entirely within `@librechat/agents` (a client-side timeout race around `sandbox.exec`), independent of `@cloudflare/sandbox`.

## Observed (production, v3.2.41)

A `bash` tool call — `cat <one source file>`, normally instant — ran **589s** and produced an empty `ExecResult`, ending with `AbortError: The operation was aborted` fired by the **host's run-level wall-clock AbortSignal** (a `processStream` `signal`), **not** by the per-tool timeout. A sibling tool call in the same batch (a schema-validation error) returned in **0.0s**, confirming the hang was the `exec()` RPC itself, not the agent loop or the model. The 589s consumed the whole run budget → degraded/partial output.

## Root cause

`src/tools/cloudflare/CloudflareSandboxExecutionEngine.ts` bounds each exec three ways; on a stalled container all three miss:

1. **In-container `timeout` wrapper** — commands are wrapped as `timeout -k 2s <n>s <cmd>` (`withInSandboxTimeout`). This only bounds a command that is *actually running inside the container*. If the container is cold/unresponsive and the command never starts, there is nothing for `timeout` to kill.
2. **`exec({ timeout })`** — `@cloudflare/sandbox`'s `BaseExecOptions.timeout`. Observed **not to fire** on a stalled RPC (the call sat ~589s with `timeout` set to ~125s). Whatever it bounds, it does not bound a container/RPC stall.
3. **AbortSignal** — `execOptions.signal` is attached only `if (ctx.sandbox.supportsExecSignal === true)`. `supportsExecSignal` is `true` only for the HTTP **bridge** runtime (`CloudflareBridgeRuntime`); for the **native Sandbox DO** it is falsy, and `@cloudflare/sandbox`'s `ExecOptions` has no `signal` field anyway. So the native-DO exec is **non-cancellable**.

Critically, there is **no client-side timeout race** around any `await ctx.sandbox.exec(...)` — 4 call sites (dist 3.2.41: lines 301, 353, 450, 462). When (1)–(3) all miss, the `await` blocks indefinitely until the host's run-level abort tears down the entire graph run.

## Why this is ours to fix

The native `exec()` can't take a signal and its `timeout` isn't reliable on a stall — but we don't need it to be. We can bound the `await` ourselves with a client-side race, fully within this package, regardless of the sandbox transport.

## Proposed fix

A client-side timeout race around each `sandbox.exec(...)`:

```ts
async function execWithClientTimeout(sandbox, command, options, timeoutMs) {
  if (!(timeoutMs > 0)) return sandbox.exec(command, options);
  let timer: ReturnType<typeof setTimeout>;
  const timeout = new Promise<never>((_, reject) => {
    timer = setTimeout(
      () => reject(new Error(`sandbox exec exceeded ${timeoutMs}ms client-side timeout`)),
      timeoutMs,
    );
  });
  try {
    return await Promise.race([sandbox.exec(command, options), timeout]);
  } finally {
    clearTimeout(timer);
  }
}
```

- Bound with the existing `outerTimeoutMs(timeoutMs)` (`timeoutMs + 5s`).
- In `createCloudflareSpawn`, on timeout call `closeOnce(124, 'SIGTERM')` so the child emits `close` and the tool returns a clear "timed out" message to the model instead of hanging.
- Keep the in-container `timeout` wrapper + `exec({timeout})` as inner bounds; the client race is the backstop for stalls they don't cover.

**Caveat:** the orphaned exec may keep running in the DO (a native-DO exec can't truly be cancelled). The race unblocks the host await; if the sandbox exposes a kill/cleanup API, call it best-effort.

## Impact

- Any cold/unresponsive container turns one tool call into a multi-minute hang that consumes the whole run budget → partial/degraded results.
- `toolExecution.cloudflare.timeoutMs` / `local.timeoutMs` are effectively ignored for the native DO under stalls.
- Surfaces more readily with fast LLM endpoints (the agent reaches tool execution quickly and often) and with intermittent container readiness.

## Affected
`src/tools/cloudflare/CloudflareSandboxExecutionEngine.ts` — `createCloudflareSpawn` (bash / local-exec path) plus the other `sandbox.exec` helpers. dist 3.2.41 call sites: lines 301, 353, 450, 462.

danny-avila:
Fixes #251.

## Problem

The native Cloudflare Sandbox Durable Object `exec()` can hang indefinitely on a stalled/cold container, and none of the existing bounds stop it:

- the in-sandbox `timeout -k 2s Ns <cmd>` wrapper only bounds a command that is *actually running* — a container that never reaches execution is unbounded;
- `exec({ timeout })` (`@cloudflare/sandbox`'s `BaseExecOptions.timeout`) is not enforced when the RPC/container itself stalls;
- `ExecOptions` has **no `signal`** (so `supportsExecSignal` is `false` for the native transport) → the exec is non-cancellable.

The **direct** exec paths (`executeCloudflareBash`, `executeCloudflareCode`, and the temp-dir cleanup) had **no client-side timeout** at all, so a stalled exec hung until the host's run-level abort — burning the entire run budget on a single tool call. Observed in production: a `cat <file>` (normally instant) ran **589s** and was only killed by the host's wall-clock `AbortSignal`.

(The spawn path used by the local coding-tool bundle already times out via `spawnLocalProcess`'s kill timer; this PR closes the gap for the direct paths.)

## Fix

Add `withClientTimeout()` — a `Promise.race` backstop around every `sandbox.exec()` — so the host await settles within `clientExecTimeoutMs()` (`outerTimeoutMs(timeoutMs) + 5s`) regardless of transport. Applied to all four `exec()` call sites in `CloudflareSandboxExecutionEngine`.

On timeout the underlying `exec` promise may keep running in the DO (a native-DO exec can't be truly cancelled), so its late settlement is swallowed to avoid an unhandled rejection.

## Test

`CloudflareSandboxExecution.test.ts`: a fake sandbox whose `exec` never resolves (simulating a stalled container) now causes `executeCloudflareCode` to **reject with a client-side timeout** instead of hanging (jest fake timers). Full suite: 22/22 green; typecheck clean.

(Note: `subagentHooks`, `hitl`, and `LocalExecutionTools` suites have pre-existing failures on `main` unrelated to this change — none import the engine touched here; verified identical failures with this change stashed.)

### DIFF
--- a/src/tools/__tests__/CloudflareSandboxExecution.test.ts
+++ b/src/tools/__tests__/CloudflareSandboxExecution.test.ts
@@ -293,6 +293,38 @@ describe('Cloudflare sandbox execution backend', () => {
     }
   });
 
+  it('aborts signal-aware execs when the client timeout fires', async () => {
+    // For signal-aware transports (e.g. the HTTP bridge), a client timeout should
+    // actually cancel the underlying exec, not just abandon it.
+    jest.useFakeTimers();
+    try {
+      let mainSignal: AbortSignal | undefined;
+      const sandbox = createRuntime({
+        supportsExecSignal: true,
+        exec: (command, options) => {
+          if (command.startsWith('rm -rf')) {
+            return Promise.resolve({ exitCode: 0, stdout: '', stderr: '' });
+          }
+          mainSignal = options?.signal;
+          return new Promise<t.CloudflareSandboxExecResult>(() => undefined);
+        },
+      });
+
+      const promise = executeCloudflareCode(
+        { lang: 'py', code: 'print("slow")' },
+        { sandbox, workspaceRoot: '/workspace', timeoutMs: 1000 }
+      );
+      const assertion = expect(promise).rejects.toThrow(/client-side timeout/);
+      await jest.advanceTimersByTimeAsync(11500);
+      await assertion;
+
+      expect(mainSignal).toBeDefined();
+      expect(mainSignal?.aborted).toBe(true);
+    } finally {
+      jest.useRealTimers();
+    }
+  });
+
   it('passes call-specific timeouts to the Cloudflare spawn wrapper', async () => {
     let execCommand = '';
     let execTimeout: number | undefined;
--- a/src/tools/__tests__/CloudflareSandboxExecution.test.ts
+++ b/src/tools/__tests__/CloudflareSandboxExecution.test.ts
@@ -365,6 +365,29 @@ describe('Cloudflare sandbox execution backend', () => {
     expect(received?.aborted).toBe(true);
   });
 
+  it('strips a caller signal for native runtimes that cannot consume it', async () => {
+    let received: t.CloudflareSandboxExecOptions | undefined;
+    const sandbox = createRuntime({
+      // no supportsExecSignal -> native DO, which cannot clone/consume a signal
+      exec: async (_command, options) => {
+        received = options;
+        return { exitCode: 0, stdout: 'ok', stderr: '' };
+      },
+    });
+    const caller = new AbortController();
+
+    await execWithClientTimeout(
+      sandbox,
+      'echo hi',
+      { cwd: '/workspace', signal: caller.signal },
+      60000,
+      'test'
+    );
+
+    expect(received).toBeDefined();
+    expect(received).not.toHaveProperty('signal');
+  });
+
   it('passes call-specific timeouts to the Cloudflare spawn wrapper', async () => {
     let execCommand = '';
     let execTimeout: number | undefined;
--- a/src/tools/__tests__/CloudflareSandboxExecution.test.ts
+++ b/src/tools/__tests__/CloudflareSandboxExecution.test.ts
@@ -257,6 +257,42 @@ describe('Cloudflare sandbox execution backend', () => {
     expect(result.timedOut).toBe(true);
   });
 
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #253, opened 2026-06-19
fix(bedrock): gate explicit prompt-cache checkpoints to Claude models

### LINKED ISSUE #13838: [Bug]: Bedrock Nova prompt caching sends invalid `toolConfig` cachePoint
### What happened?

Nova models started failing after the Bedrock prompt-caching change in:

`c9ee0f138` - `feat: Add Bedrock Prompt Caching Support (#8271)`

Before this change, Nova requests worked with the same model configuration. After this change, Nova model strings are auto-defaulted to `promptCache: true`:

```ts
// packages/data-provider/src/bedrock.ts
if (
  typeof typedData.model === 'string' &&
  (typedData.model.includes('claude') || typedData.model.includes('nova'))
) {
  if (typedData.promptCache === undefined) {
    typedData.promptCache = true;
  }
}
```

That then causes `@librechat/agents` to inject a cache point into `toolConfig.tools`:

```ts
const toolConfig =
  this.promptCache === true
    ? insertBedrockToolCachePoint(baseParams.toolConfig, true)
    : baseParams.toolConfig;
```

AWS Bedrock rejects this for Nova:

```text
Malformed input request: #/toolConfig/tools/0: extraneous key [cachePoint] is not permitted
```

### Expected

Nova prompt caching should not generate an invalid `toolConfig.tools[].cachePoint` request. Either Nova should not be defaulted to `promptCache: true`, or the cache point placement should be model-family aware.

### Workaround

Set `promptCache: false` for Nova model specs.
```

### Version Information

268fcbb78d234ecd271f88496a4c554617eedcb0

### Steps to Reproduce

start a chat with a nova model

### What browsers are you seeing the problem on?

_No response_

### Relevant log output

```shell
na
```

### Screenshots

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

danny-avila:
## Problem

Fixes [danny-avila/LibreChat#13838](https://github.com/danny-avila/LibreChat/issues/13838). Amazon Nova requests failed with:

```
ValidationException: Malformed input request: #/toolConfig/tools/0: extraneous key [cachePoint] is not permitted
```

LibreChat auto-defaults Nova to `promptCache: true`, and the SDK then injected a `cachePoint` into `toolConfig.tools`, which Nova rejects.

## What Nova actually supports (live-probed, Converse, `us.amazon.nova-lite-v1:0`)

| cachePoint location | result |
| --- | --- |
| `toolConfig.tools` | ❌ 400 `extraneous key [cachePoint] is not permitted` |
| `messages` content | ✅ HTTP 200, `cacheWriteInputTokens=240` |
| `system` | ✅ HTTP 200, `cacheWriteInputTokens=241` |

So **only the tool checkpoint is Claude-only** — Nova caches system/messages fine (and it's real, paid caching). An earlier revision of this PR gated *all* Bedrock caching to Claude; that was too aggressive (it threw away valid Nova message/system caching), so it's been narrowed to the tool path only.

## Fix

Add `supportsBedrockToolCache(model)` (Claude-only) and gate **only** the tool cache point on it:

- **`bedrock/index.ts`** `invocationParams` — skip the `toolConfig` cache point on non-Claude models. Gated on a `cacheModelId` **captured at construction**, so it survives the `this.model` → application-inference-profile ARN swap during generation.
- **`Graph.ts`** — skip Bedrock tool *marking* on non-Claude models. The message/system tail cache (`bedrockPromptCacheEnabled`) stays gated on `promptCache` alone.
- **`AgentContext.ts`** summarization path — stays gated on `promptCache` alone.

Net: Nova keeps system/message caching; only the unsupported tool checkpoint is suppressed. Capability gate (key hard-rejected), distinct from the 1h/5m TTL value which downgrades gracefully on Claude.

## Testing

- **Live**: probe above — tool 400, message/system 200 with cache writes.
- **Unit**: `supportsBedrockToolCache` (Claude→true, Nova/Llama/Mistral/empty→false); `invocationParams` (Nova → no tool cachePoint; Claude behind an inference-profile ARN → still cached, swap-immune); existing Claude prompt-cache tests specify a Claude model.

`tsc --noEmit` clean; bedrock + messages + agents suites green.

### DIFF
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -846,6 +846,8 @@ export class AgentContext {
     const bedrockOptions = this.clientOptions as
       | t.BedrockAnthropicClientOptions
       | undefined;
+    // Nova accepts system/message cachePoints (only the tool checkpoint is
+    // Claude-only), so this is gated on promptCache alone.
     return bedrockOptions?.promptCache === true;
   }
 
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -1441,16 +1441,21 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
           agentContext.clientOptions as
             | t.BedrockAnthropicClientOptions
             | undefined
-        )?.promptCache === true &&
-        supportsBedrockToolCache(
-          (agentContext.clientOptions as { model?: string } | undefined)?.model
-        )
+        )?.promptCache === true
       ) {
-        toolsForBinding =
-          partitionAndMarkBedrockToolCache(
-            rawToolsForBinding,
-            makeIsDeferred(agentContext.toolDefinitions)
-          ) ?? rawToolsForBinding;
+        const bedrockModel = (
+          agentContext.clientOptions as { model?: string } | undefined
+        )?.model;
+        // An omitted model falls back to LangChain's default Claude model (which
+        // supports tool caching); only an explicit non-Claude model (e.g. Nova)
+        // skips tool marking so its stray marker never leaks into toolConfig.
+        if (bedrockModel == null || supportsBedrockToolCache(bedrockModel)) {
+          toolsForBinding =
+            partitionAndMarkBedrockToolCache(
+              rawToolsForBinding,
+              makeIsDeferred(agentContext.toolDefinitions)
+            ) ?? rawToolsForBinding;
+        }
       }
 
       let model =
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -28,6 +28,7 @@ import {
   syncBudgetDerivedFields,
   addTailCacheControl,
   resolvePromptCacheTtl,
+  supportsBedrockToolCache,
   getMessageId,
   makeIsDeferred,
   partitionAndMarkAnthropicToolCache,
@@ -1440,7 +1441,10 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
           agentContext.clientOptions as
             | t.BedrockAnthropicClientOptions
             | undefined
-        )?.promptCache === true
+        )?.promptCache === true &&
+        supportsBedrockToolCache(
+          (agentContext.clientOptions as { model?: string } | undefined)?.model
+        )
       ) {
         toolsForBinding =
           partitionAndMarkBedrockToolCache(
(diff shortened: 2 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Both addressed in `d9305c7` — same root cause (an omitted `model` falls back to LangChain's default Claude model, which my gates treated as unsupported). Verified the default is a Claude model.

**P2 — Use the superclass default model for cache gating (`index.ts`).** Fixed: `cacheModelId = fields?.model ?? this.model`. `super(fields)` initializes `this.model` to the default Claude model when `fields.model` is omitted, so the captured id is now that default instead of `''`. (Still swap-immune — captured at construction, before the inference-profile ARN swap.)

**P2 — Account for the default Claude model before graph marking (`Graph.ts`).** Fixed: the gate now treats an omitted `clientOptions.model` as the default Claude model — `bedrockModel == null || supportsBedrockToolCache(bedrockModel)` — so the default path still gets tool marking/partitioning; only an explicit non-Claude model (Nova) skips it. (The marker is a string key that would otherwise leak into `toolConfig`, so skipping the *marking* for Nova — not just the materialization — is the right level.)

Added a regression test: omitted model + `promptCache: true` + tools → tool cache point still applied (and `model.model` matches `/claude/i`). tsc clean; bedrock + messages + agents suites green (256 passing).

@codex review

danny-avila:
Pushed : clamp the extended 1h TTL to 5m on non-Claude Bedrock models. Nova hard-rejects `ttl:'1h'` (`Extended TTL prompt caching is only supported for Anthropic models`, verified live), so the 1h default was breaking Nova's message/system caching that worked at 5m. New `resolveBedrockPromptCacheTtl(ttl, model)` keeps 1h for Claude/default and clamps non-Claude to 5m.

@codex review

danny-avila:
This was reviewed against a stale merge preview (`3941c41` — the pre-amend commit message, content from before the TTL fix). It's already implemented at HEAD `31648fe`: all three Bedrock cachePoint sites resolve the TTL with the model via `resolveBedrockPromptCacheTtl(ttl, model)`, which clamps non-Claude models to 5m.

Verified **end-to-end** through the built SDK against live Bedrock:

```
[NOVA]   resolveBedrockPromptCacheTtl(undefined, nova)  -> '5m'
         addBedrockTailCacheControl -> cachePoint {"type":"default"}
         AWS Converse -> HTTP 200, cacheWriteInputTokens=240
[CLAUDE] resolveBedrockPromptCacheTtl(undefined, claude) -> '1h'
         cachePoint {"type":"default","ttl":"1h"}
         AWS Converse -> HTTP 200
```

So Nova message/system caching works at 5m; only Claude gets the extended 1h TTL.

@codex review

PR #254, opened 2026-06-19
test: cover custom OpenAI-compatible vLLM reasoning + qwen3_coder tool calls

danny-avila:
## Summary

Adds a regression spec for a generic `custom` OpenAI-compatible endpoint (provider `openai`, **default** `reasoningKey`, non-standard model name) that:

1. **Streams reasoning in the modern vLLM `reasoning` field** — `reasoning_content` is `null` throughout. Current vLLM emits CoT only in `reasoning` (`DeltaMessage.reasoning` / `ChatMessage.reasoning`); `reasoning_content` survives just as a deprecated *input* alias that vLLM renames to `reasoning` ([RFC vllm#27755](https://github.com/vllm-project/vllm/issues/27755)).
2. **Streams `tool_calls` with fragmented arguments** (`--tool-call-parser qwen3_coder`): id/name on the first delta, args split across subsequent role-less continuation deltas.

The spec locks in that:
- `delta.reasoning` surfaces as `think` content (the Thoughts block) ahead of the answer `text`, through the real `Run`/`StandardGraph` streaming path; and
- the fragmented streamed tool call accumulates into a structured `get_weather({"location":"Berlin"})`.

Both exercise real SDK code paths (graph dispatch + the actual per-delta `_convertCompletionsDeltaToBaseMessageChunk` conversion); the only mocked boundary is the upstream HTTP stream.

## Motivation

Wire shapes mirror the captures in LibreChat discussion [#13849](https://github.com/danny-avila/LibreChat/discussions/13849), which reported reasoning/tool-calling breaking on a custom vLLM endpoint after upgrading LibreChat. Replaying the exact deltas shows both render/parse correctly on current code, so this guards the `reasoning`-field normalization and the fragmented-`tool_calls` accumulation against future regressions.

## Test plan

- [x] `npx jest src/specs/vllm-reasoning-toolcalls.test.ts` — 2 passed
- [x] `npx eslint src/specs/vllm-reasoning-toolcalls.test.ts` — clean
- [x] `npx tsc --noEmit` — clean

### DIFF
--- a/src/specs/vllm-reasoning-toolcalls.test.ts
+++ b/src/specs/vllm-reasoning-toolcalls.test.ts
@@ -1,0 +1,251 @@
+/**
+ * Regression coverage for a generic `custom` OpenAI-compatible endpoint
+ * (provider `openai`, default `reasoningKey`, non-standard model name) that
+ * streams reasoning in the modern vLLM `reasoning` field — `reasoning_content`
+ * is `null` throughout — and streams `tool_calls` with fragmented arguments
+ * (vLLM `--reasoning-parser qwen3` + `--tool-call-parser qwen3_coder`).
+ *
+ * The wire shapes mirror the captures in LibreChat discussion #13849:
+ *   - reasoning must surface as `think` content (the Thoughts block), and
+ *   - the fragmented streamed tool call must accumulate into a structured call.
+ */
+import { AIMessageChunk, HumanMessage } from '@langchain/core/messages';
+import type { OpenAIClient } from '@langchain/openai';
+import type * as t from '@/types';
+import { ContentTypes, GraphEvents, Providers } from '@/common';
+import { createContentAggregator } from '@/stream';
+import { ChatOpenAI } from '@/llm/openai';
+import { Run } from '@/run';
+
+type StreamingCompletionBackedModel = {
+  completions: {
+    completionWithRetry: () => Promise<
+      AsyncIterable<OpenAIClient.Chat.Completions.ChatCompletionChunk>
+    >;
+  };
+};
+
+type DeltaConverterModel = {
+  completions: {
+    _convertCompletionsDeltaToBaseMessageChunk(
+      delta: Record<string, unknown>,
+      rawResponse: Record<string, unknown>,
+      defaultRole?: string
+    ): AIMessageChunk;
+  };
+};
+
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

danny-avila:
Addressed all three Codex findings in 7d602c5:

1. **Non-OpenAI `baseURL` (P2, line 134).** Correct — with an unset baseURL `isOfficialOpenAIBaseURL` returns `true`, so the spec was exercising the official-OpenAI sequential-seal path, not the custom endpoint. Both `ChatOpenAI` instances now use `configuration: { baseURL: 'http://vllm.internal:8000/v1' }` so they stay on the custom-endpoint final-signal path.
2. **Tool calls through the graph (P2, line 225).** Correct — the converter-only test couldn't catch a graph-path regression. The tool-call test now runs through `Run.processStream` with a real `get_weather` tool and asserts (a) the streamed fragments are dispatched into a `TOOL_CALLS` run step naming `get_weather`, (b) the tool executes with the assembled args `{ location: 'Berlin' }`, and (c) a tool-result message + final answer come back (model invoked twice).
3. **Type the converter inputs (P2, line 32).** Done — `Record<string, unknown>` is replaced with typed `ChatCompletionChunk` / a `CompletionDelta` extended with the vLLM `reasoning`/`reasoning_content` fields, per AGENTS.md.

@codex review

danny-avila:
Addressed the round-2 finding in 7344635.

**Capture streamed tool-call deltas (P2).** Correct and a sharp catch — the `TOOL_CALLS` `ON_RUN_STEP` can be produced from the final assembled `AIMessage` via `handleToolCalls` ([stream.ts:1412](../blob/main/src/stream.ts#L1412)) even if the streamed-delta path `handleToolCallChunks` ([stream.ts:1484](../blob/main/src/stream.ts#L1484)) regresses, so asserting only on the step name would miss the streaming-UI regression — which is the actual reported symptom.

The tool-call test now registers an `ON_RUN_STEP_DELTA` handler, collects the `tool_calls[].args` fragments dispatched by `handleToolCallChunks`, and asserts they assemble to `{"location": "Berlin"}`. That fails on empty (no streamed fragments), so it genuinely guards the streamed path in addition to the existing final-assembly + tool-execution assertions.

Tests pass, eslint + tsc clean.

@codex review

PR #255, opened 2026-06-19
fix(cloudflare): client-side timeout on native sandbox file-IO RPCs

danny-avila:
## Summary

Sequel to #252 (which bounded `sandbox.exec()`). The native-DO **file-IO** RPCs were still unbounded.

`createCloudflareWorkspaceFS`'s `readFile` / `writeFile` / `stat` / `readdir` / `mkdir` / `unlink` / `open` call `sandbox.readFile()` / `writeFile()` / `listFiles()` / `mkdir()` / `deleteFile()` directly — the same uncancellable native Durable Object transport that #251/#252 fixed for exec: no `signal` on `ExecOptions`, and `timeout` is not reliably enforced when the container/RPC itself stalls. So a stalled or cold container hangs the host `await` on a **single file read** until the run-level abort, burning the whole run budget on one tool call.

## Why now

Observed live in issue-triage: a `read_file` stalled **~552s** before the wall-clock budget killed the run. In the same parallel tool batch, an `execute_bash` was correctly bounded at ~130s (by #252) — only the `read_file` hung, because the FS path has no client-side backstop. The earlier "prefer `read_file` over `bash`" guidance had ironically shifted traffic off the now-bounded exec path onto the still-unbounded FS path.

## Change

- New `clientFsTimeoutMs(timeoutMs) = timeoutMs + 5000`. File IO has no in-sandbox `timeout(1)` layer to honor (unlike exec, which gets `outerTimeoutMs` + client headroom), so a single margin over the configured tool timeout is the right backstop. Reads are byte-capped, so a normal read completes well within it; a stalled container can't outlast it.
- Wrap **every** native FS RPC in `createCloudflareWorkspaceFS` — plus the `listFiles` inside `findChildInfo` — with the existing `withClientTimeout` helper (same Promise.race backstop the exec sites use; the native DO RPC can't be truly cancelled, so the late settlement is swallowed).

## Tests

- Stalled `readFile` → rejects with a client-side timeout (the live `read_file` regression).
- Stalled `listFiles` → rejects with a client-side timeout (covers `readdir`/`stat`).
- A fast native FS RPC resolves and leaves **no dangling backstop timer**.

Cloudflare suite **28/28**, `tsc` + `eslint` clean.

Refs #251, #252.

### DIFF
--- a/src/tools/__tests__/CloudflareSandboxExecution.test.ts
+++ b/src/tools/__tests__/CloudflareSandboxExecution.test.ts
@@ -507,6 +507,36 @@ describe('Cloudflare sandbox execution backend', () => {
     }
   });
 
+  it('still cleans up the temp dir when execute_code setup (writeFile) times out', async () => {
+    // The setup RPCs are inside the try, so a stalled writeFile still triggers
+    // the finally cleanup — otherwise the late (uncancellable) write leaves an
+    // orphaned .lc-exec/<uuid> dir behind on every cold-container failure.
+    jest.useFakeTimers();
+    try {
+      const execCommands: string[] = [];
+      const sandbox = createRuntime({
+        mkdir: async () => ({ ok: true }),
+        writeFile: () => new Promise<{ ok: true }>(() => undefined), // stalls
+        exec: async (command) => {
+          execCommands.push(command);
+          return { exitCode: 0, stdout: '', stderr: '' };
+        },
+      });
+
+      const promise = executeCloudflareCode(
+        { lang: 'py', code: 'print("hi")' },
+        { sandbox, workspaceRoot: '/workspace', timeoutMs: 1000 }
+      ).catch((e: unknown) => e);
+      await jest.advanceTimersByTimeAsync(6500);
+      const settled = await promise;
+      expect(isWorkspaceClientTimeoutError(settled)).toBe(true);
+      // Cleanup must have been issued despite the setup failure.
+      expect(execCommands.some((c) => c.startsWith('rm -rf'))).toBe(true);
+    } finally {
+      jest.useRealTimers();
+    }
+  });
+
   it('rejects with a client-side timeout when sandbox listFiles stalls', async () => {
     jest.useFakeTimers();
     try {
--- a/src/tools/__tests__/CloudflareSandboxExecution.test.ts
+++ b/src/tools/__tests__/CloudflareSandboxExecution.test.ts
@@ -388,6 +388,81 @@ describe('Cloudflare sandbox execution backend', () => {
     expect(received).not.toHaveProperty('signal');
   });
 
+  it('rejects with a client-side timeout when sandbox readFile stalls', async () => {
+    // The native-DO file-IO RPCs (readFile/writeFile/listFiles/...) have the same
+    // stall hazard exec() does: no signal, no enforced timeout. A cold/unresponsive
+    // container otherwise hangs the host await until the run-level abort, burning
+    // the whole budget on one read (observed live: a `read_file` stalled ~552s).
+    jest.useFakeTimers();
+    try {
+      let readCalls = 0;
+      const fs = createCloudflareWorkspaceFS({
+        workspaceRoot: '/workspace',
+        timeoutMs: 1000,
+        sandbox: createRuntime({
+          readFile: () => {
+            readCalls += 1;
+            return new Promise<string>(() => undefined);
+          },
+        }),
+      });
+
+      const promise = (fs.readFile as (p: string) => Promise<unknown>)(
+        '/workspace/a.txt'
+      );
+      const assertion = expect(promise).rejects.toThrow(/client-side timeout/);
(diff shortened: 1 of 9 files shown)


danny-avila:
@codex review

danny-avila:
Thanks — addressed in `0e44ed0`:

- **P1 (timeouts masquerading as missing files):** the backstop now throws a distinguishable `WorkspaceClientTimeoutError` (`code: WORKSPACE_CLIENT_TIMEOUT`, exported from `workspaceFS.ts`). The two ENOENT-only fallbacks rethrow it — the `write_file` pre-read in `LocalCodingTools` and `FileCheckpointer.stat` — so a stalled read/stat can no longer be mistaken for an absent file (which would overwrite, or snapshot-as-absent and delete on rewind).
- **P2 (streamed-read drain):** the race now wraps `readFile` + `normalizeReadFileContent` together at all three read sites (`readFile`/`open`/`stat`-fallback), so a `{ content: ReadableStream }` that stalls mid-drain can't outlast the timer.
- **P2 (`execute_code` setup):** `executeCloudflareCode`'s pre-exec `ctx.sandbox.mkdir`/`writeFile` are now bounded with the same backstop.
- **P2 (mutating-RPC late write):** not addressed in this PR. A timed-out `writeFile`/`mkdir`/`deleteFile` abandons the host await but can't cancel the native DO RPC, so a late write may still land — the same uncancellable-transport property #252 accepted for `exec()` ("late settlement is swallowed"). Poisoning the run on a mutating timeout is a larger behavior change; flagging it for a separate decision rather than folding it into this focused backstop.

New tests: distinguishable-error assertion on the `readFile` stall, streamed-drain stall, `execute_code` mkdir-setup stall, and `FileCheckpointer` rethrowing a stat timeout instead of recording absent. `CloudflareSandboxExecution` 30/30; local-tool + checkpointer suites 306 passed; tsc + eslint clean.

@codex review

danny-avila:
Addressed all three in `af57131` — same principle (a typed FS timeout must never be silently absorbed):

- **stat directory-probe (engine):** the `listFiles` probe catch now rethrows `WorkspaceClientTimeoutError` instead of falling through to the `readFile` branch, so a stalled stat surfaces immediately rather than waiting through a second full backstop (~2× the timeout).
- **fallback walker (`grep_search`/`glob_search`):** the `readdir` catch rethrows the timeout instead of skipping the directory (which silently corrupted results to "no matches").
- **fallback grep scanner:** both the per-file `stat` and `readFile` catches rethrow the timeout instead of skipping the file. The outer fallback handler already rethrows non-`FallbackGrepError`, so these surface as a real failure.

Added an engine test asserting the stat probe timeout rethrows and does **not** fall through to `readFile`. `CloudflareSandboxExecution` 31/31; local-tool/checkpointer suites 275 passed; tsc + eslint clean.

@codex review

danny-avila:
Addressed both in `bf7b0b1`, and did a full sweep of the remaining `WorkspaceFS`-error swallow sites so we converge rather than round-trip:

**Probes (were producing wrong results):**
- `CompileCheckTool.pathExists` — stat timeout no longer returns `false` (wrong toolchain).
- `CompileCheckTool` `package.json` / `pyproject.toml` reads — timeout no longer substitutes `''` (missed typescript/mypy detection).
- `syntaxCheck` `jsonCheck` read — stalled read no longer returns `ok: true` (false syntax-gate pass).
- `syntaxCheck` `runPostEditSyntaxCheck` — checker timeout no longer collapses to `null` (write would pass without a real gate).

**Rollback (were claiming false success):**
- `revertStrictWrite` — a timed-out revert no longer lets the caller report "reverted to pre-write state" while the rejected bytes remain.
- `FileCheckpointerImpl.rewind` — a timed-out restore write *or* delete no longer counts the path as restored; rewind surfaces the timeout.

Non-timeout failures keep their existing benign behavior at every site. Added a rewind test (restore-write timeout → rejects, not silent success). Affected suites green (181 + 127 passed); tsc + eslint clean.

The invariant across all of this: a `WorkspaceClientTimeoutError` (stalled native-DO RPC) must always surface — never absorbed as absent / empty / skip / no-match / reverted.

@codex review

danny-avila:
@codex review

danny-avila:
Both fixed in `8d616a5`:

- **execute_code temp-dir leak:** moved the client-bounded `mkdir`/`writeFile` setup *inside* the `try`, so the existing detached `finally` cleanup removes `.lc-exec/<uuid>` even when setup times out (the uncancellable write may land late on a cold container).
- **strict-mode revert bypass:** wrapped `maybeRunSyntaxCheck` in `write_file`/`edit_file` so a stalled validation read routes through `revertStrictWrite` (in strict mode) before surfacing — best-effort, since the revert can itself time out on the same stalled container.

Tests: `execute_code` still issues `rm -rf` cleanup when the setup `writeFile` stalls; `write_file` reverts (unlinks the new file) when the strict validation read times out. Affected suites 183 passed; tsc + eslint clean.

@codex review

danny-avila:
@codex review

PR #256, opened 2026-06-19
fix(bedrock): clamp the extended 1h TTL to 5m on non-Claude models

danny-avila:
## Problem

Follow-up to [#253](https://github.com/danny-avila/agents/pull/253), which fixed the Nova **tool** cachePoint but landed before this change. The remaining issue: the extended **1h** prompt-cache TTL is Anthropic-only on Bedrock. Non-Claude models reject it —

```
ValidationException: Extended TTL prompt caching is only supported for Anthropic models
```

— so the 1h default (from the prompt-cache TTL work) breaks Nova's **system/message** caching, which #253 deliberately keeps enabled. Nova caching worked before at 5m; the 1h default regressed it to a 400.

## Fix

Add `resolveBedrockPromptCacheTtl(ttl, model)`:
- **Claude** (and the omitted-model default, which resolves to Claude) → keep the `1h` default / configured ttl. Claude downgrades an unsupported 1h to 5m server-side.
- **Non-Claude (Nova, Llama, Mistral, …)** → clamp to `5m`, even when `1h` is explicitly configured.

Wired into the three Bedrock cachePoint TTL sites: `bedrock/index.ts` (tool — already Claude-gated), `Graph.ts` (message tail), `AgentContext.ts` (summary). Shares one `isBedrockAnthropicModel` predicate with `supportsBedrockToolCache`.

## Testing

**Live, end-to-end** through the built SDK (`resolveBedrockPromptCacheTtl` + `addBedrockTailCacheControl` → AWS Converse):

```
[NOVA]   resolved ttl 5m -> cachePoint {"type":"default"}        -> HTTP 200, cacheWrite=240
[CLAUDE] resolved ttl 1h -> cachePoint {"type":"default","ttl":"1h"} -> HTTP 200
```

**Unit:** `resolveBedrockPromptCacheTtl` — Claude defaults to 1h / honors explicit ttl; Nova + others clamp to 5m even with explicit 1h; omitted model → 1h.

`tsc --noEmit` clean; cache + bedrock + agents suites green (260 passing).

Fixes the extended-TTL half of danny-avila/LibreChat#13838.

### DIFF
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -14,6 +14,7 @@ import {
   buildAnthropicCacheControl,
   buildBedrockCachePoint,
   resolvePromptCacheTtl,
+  resolveBedrockPromptCacheTtl,
   cloneMessage,
   type PromptCacheTtl,
 } from '@/messages/cache';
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #258, opened 2026-06-19
fix(reducer): skip null/undefined entries before coerceMessageLikeToMessage

cosmic-fire-eng:
`messagesStateReducer` maps both input arrays through `coerceMessageLikeToMessage` without first checking for null/undefined entries. When a provider emits an empty or partial stream chunk, that entry arrives as `undefined`, and `coerceMessageLikeToMessage` throws `Cannot read properties of undefined (reading 'role')`, which crashes the LangGraph run.

We hit this consistently on tool-heavy runs with certain providers. Filtering out null/undefined before coercion is a minimal, defensive fix that preserves message order and leaves all existing behavior (id assignment, REMOVE_ALL handling, merge/dedupe) untouched.

Refs LibreChat Discussion #12284.

### DIFF
--- a/src/messages/reducer.spec.ts
+++ b/src/messages/reducer.spec.ts
@@ -1,0 +1,32 @@
+import { HumanMessage } from '@langchain/core/messages';
+import { messagesStateReducer } from './reducer';
+
+describe('messagesStateReducer', () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #263, opened 2026-06-23
feat(search): add Keenable as a search provider

ilya-bogin-keenable:
Adds Keenable as a first-class search provider, modeled on the existing Tavily provider.

Context: this comes out of danny-avila/LibreChat#13805, where @danny-avila pointed out that the right home for Keenable is a provider here rather than a standalone tool in LibreChat, so it inherits the shared scraping/reranking/highlighting pipeline. This is that move.

What's here:
- `keenable-search.ts` — `createKeenableAPI`, maps Keenable's `/v1/search` results into `organic` sources. The rest of the pipeline (scraping, reranking, highlights, formatting) is reused unchanged.
- `types.ts` — `'keenable'` added to `SearchProvider`; `KeenableSearchOptions` plus payload/response types.
- `search.ts` / `tool.ts` — wired into `createSearchAPI` and `createSearchTool`. No video search for Keenable, same as Tavily.
- `keenable.test.ts` — keyless vs authenticated endpoint, field mapping, `site` filter, client-side result count, error handling.

Notes:
- Keyless by default: with no key it hits the public endpoint; `X-API-Key` is only sent when a key is set. A key just lifts rate limits.
- The endpoint has no result-count parameter, so the count is applied client-side.
- LibreChat-side config (the `SearchProviders` enum, schema, and auth wiring) is a separate follow-up PR once this is released.

Tests pass, lint and typecheck clean.

### DIFF
--- a/src/tools/search/keenable-search.ts
+++ b/src/tools/search/keenable-search.ts
@@ -1,0 +1,86 @@
+import axios from 'axios';
+import type * as t from './types';
+
+const DEFAULT_KEENABLE_TIMEOUT = 15000;
+
+/** Authenticated and keyless endpoints. Keenable works without an API key by
+ * falling back to the public endpoint; a key only lifts rate limits. */
+const KEENABLE_DEFAULT_API_URL = 'https://api.keenable.ai/v1/search';
+const KEENABLE_PUBLIC_API_URL = 'https://api.keenable.ai/v1/search/public';
+
+export const createKeenableAPI = (
+  apiKey?: string,
+  apiUrl?: string,
+  options?: t.KeenableSearchOptions
+): {
+  getSources: (params: t.GetSourcesParams) => Promise<t.SearchResult>;
+} => {
+  const resolvedKey = apiKey ?? process.env.KEENABLE_API_KEY;
+  const hasKey = resolvedKey != null && resolvedKey !== '';
+  const timeout = options?.timeout ?? DEFAULT_KEENABLE_TIMEOUT;
+  const resolvedUrl =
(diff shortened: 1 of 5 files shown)


ilya-bogin-keenable:
@danny-avila could you please take a look? Thanks!

ilya-bogin-keenable:
The failing checks are the provider E2E specs (`anthropic.simple`, `openai.simple`, `cache.simple`, `summarization`). They need `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`, which Actions doesn't expose to PRs from forks, so they fail with "Could not resolve authentication method" no matter what the change is. `src/tools/search/keenable.test.ts` passes, and lint, typecheck and circular-deps are green. This PR only touches `src/tools/search/`.

danny-avila:
Thanks @ilya-bogin-keenable, the provider is solid — I live-tested both the public and authenticated endpoints and the full search pipeline with it, and everything checks out.

One fix was needed: the web-search `date` parameter was ignored, so date-bounded searches silently returned unbounded results. I have a commit mapping `DATE_RANGE` to Keenable's `published_after`, but GitHub does not allow maintainer pushes to organization-owned forks (even with "allow edits by maintainers" enabled), so I could not push it here.

To keep this moving I have continued it in https://github.com/danny-avila/agents/pull/285 with your commit preserved as-is (authorship intact), plus the date-filter fix and a rebase onto current main. Closing this one in favor of that PR — thanks again for the contribution, and the LibreChat-side follow-up you mentioned is still welcome once it ships.

PR #272, opened 2026-06-24
feat(run): forward update + goto on resume (langgraph 1.4.5)

danny-avila:
## Summary

Extends the SDK's `Run.resume()` to forward `update` + `goto` into the resume `Command` (langgraph 1.4.5), so a human-in-the-loop approval can **commit a state edit and reroute in the same superstep** (one checkpoint, no flicker) — not just resume.

- `resume()` gains an optional 4th param `commandOptions?: Pick<ConstructorParameters<typeof Command>[0], 'update' | 'goto'>`, threaded into `new Command({ resume, update?, goto? })` (`src/run.ts`).
- **Backward-compatible:** existing `resume(value, config, streamOptions?)` callers are untouched — `update` is omitted and `goto` stays `[]` unless the caller sets them.
- This is the concrete HITL-resume capability flagged in the durable-substrate design (#271): `resume()` previously only built `Command({ resume })`.

Self-contained in `run.ts` — no overlap with the in-flight `runtime.state` (ToolNode/MultiAgentGraph) or stream-dispatch work, and doesn't depend on the durable-checkpointer decisions.

## Testing

`npm run build` clean. New `Run.create`-harness test asserts the resume `Command` carries `update` + (array-normalized) `goto`, and that omitting `commandOptions` leaves `update` undefined / `goto` `[]`. Full deterministic suite green (135 suites / 2481 tests).

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -1139,6 +1139,20 @@ export class Run<_T extends t.BaseGraphState> {
     return cp == null ? 0 : cp.rewind();
   }
 
+  /**
+   * Resume an interrupted run. `commandOptions` forwards langgraph 1.4.5
+   * `Command` fields applied together with `resume` in one superstep:
+   * - `update`: channel updates committed at the resume point. On a *rebuilt*
+   *   Run (new instance + durable checkpointer), `update.messages` are the first
+   *   write the fresh wrapper sees, so they seed the `getRunMessages()` /
+   *   `returnContent` baseline and are excluded from them (still committed to the
+   *   checkpoint). Hosts that rebuild + inject messages should persist them
+   *   directly or read from `getState`. Unreachable without a durable checkpointer.
+   * - `goto`: a *dynamic* edge that does not cancel static `addEdge` routes. On
+   *   the built-in standard graph the fixed `toolNode -> agentNode/END` edge still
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
On the P2 about input `Command` fields: langgraph 1.4.5 **does** honor `update`/`goto` on an input (resume) Command. The input Command runs through `mapCommand` (`pregel/io.js:30`), which emits writes for `resume` (L48-61), `update` (L62-74), and `goto` (L32-47); `pregel/loop.js:648` calls `mapCommand(this.input, ...)` when `isCommand(this.input)`, and `loop.js:662,679` has a dedicated `isCommandUpdateOrGoto` input path.

Fair point that the prior test only inspected the constructed object. Added an executing test (`hitl.test.ts`) that interrupts on a tool, resumes via `Run.resume(decision, config, undefined, { update: { messages } })`, then reads the committed state back with `graph.getState` and asserts the injected message is present and the resume completed. It exercises the real `streamEvents` path. Full HITL suite green (60/60).

danny-avila:
@codex review

PR #273, opened 2026-06-24
feat: forward runtime.state to tools (langgraph 1.4.1), drop getCurrentTaskInput

danny-avila:
## Summary

Adopts langgraph 1.4.1's `ToolRuntime.state` so tools read graph state off their second argument, replacing the deprecated `getCurrentTaskInput()` (which relies on `node:async_hooks` and is browser-incompatible). First consumer of the new langgraph 1.4 APIs.

- **`ToolNode`** (our fork extends `RunnableCallable`, *not* langgraph's prebuilt ToolNode, so it didn't forward state): now threads the run input (graph state) through `run() → runTool → tool.invoke`, building langgraph's exact runtime shape `{ ...config, state, toolCallId, config, context, store, writer }` as the tool's 2nd arg.
- **`MultiAgentGraph`** handoff tools (2) and the `handoff-test` dev script: `getCurrentTaskInput()` → `runtime.state`. **Zero `getCurrentTaskInput()` calls remain** in `src/`.
- New `ToolNode.runtimeState.test.ts` proves forwarding for message-state, array, and `Send`-de-enveloped inputs (+ `toolCallId`).

### Deliberately NOT changed (verified independent — removing them would be wrong)
- `run.ts` `__pregel_scratchpad.currentTaskInput` leak cleanup — langgraph populates `currentTaskInput` **unconditionally** for every task (`pregel/algo.js`), regardless of `getCurrentTaskInput` usage, so the cleanup stays.
- `ToolNode` `interrupt()` `AsyncLocalStorage` shim — `interrupt()` hard-requires the ALS frame (`interrupt.js` throws without it); unrelated to state access.

## Testing

`npm run build` + full `tsc --noEmit` clean; new forwarding test green; multi-agent/handoff/ToolNode suites green; full deterministic suite green (135 suites / 2454 tests — one unrelated pre-existing parallel-worker flake that passes on isolation/re-run). Live suites run in CI.

### DIFF
--- a/src/graphs/MultiAgentGraph.ts
+++ b/src/graphs/MultiAgentGraph.ts
@@ -12,12 +12,11 @@ import {
   Command,
   StateGraph,
   Annotation,
-  getCurrentTaskInput,
   messagesStateReducer,
 } from '@langchain/langgraph';
 import type { BaseMessage, AIMessageChunk } from '@langchain/core/messages';
 import type { LangGraphRunnableConfig } from '@langchain/langgraph';
-import type { ToolRunnableConfig } from '@langchain/core/tools';
+import type { ToolRuntime } from '@langchain/core/tools';
 import type * as t from '@/types';
 import { StandardGraph } from './Graph';
 import { Constants } from '@/common';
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #276, opened 2026-06-26
feat: make mainExpandBy and separatorExpandBy configurable options

Theroxenes:
## Summary

Web search highlights are expanded on both sides by `expandHighlights` to give
the model surrounding context, using hardcoded `mainExpandBy = 300` /
`separatorExpandBy = 150` (chars per side). This expansion can be the dominant 
contributor to web search payload size.

Measured on a real search (local SearXNG + Firecrawl-compatible scraper +
CrossEncoder reranker), per source, holding the reranked chunks constant and
varying only the expansion params:

| Expansion (main/separator) | Post-expansion bytes/source | Ratio vs. reranked chunks |
| -------------------------- | --------------------------- | ------------------------- |
| 300 / 150 (current default) | ~4,200                      | ~6.7×                     |
| 100 / 50                    | ~1,700                      | ~2.8×                     |

A ~58% per-source reduction, with highlight coherence roughly preserved.

This PR exposes the two parameters via config, following the existing
`topResults` / `strategies` / `safeSearch` convention (config-only — no env
var). Defaults are unchanged, so the change is non-breaking.

`expandHighlights` already accepts these parameters with the correct defaults;
this PR simply threads config values to the existing call site rather than
adding new machinery.

## Changes

- `types.ts`: add `mainExpandBy` / `separatorExpandBy` to `ProcessSourcesConfig`
  (inherited by `SearchToolConfig`).
- `tool.ts`: destructure both from config with defaults (300/150) and pass them
  to the `expandHighlights` call.
- `source-processing.test.ts`: add coverage for default expansion, smaller →
  shorter highlights, larger → longer, zero → chunk-only, and the
  processor→expand path.

## Change Type

## Testing

- `npx jest src/tools/search` — all existing + new tests pass.
- `npx tsc --noEmit`, `npx eslint src/tools/search/` clean.
- Verified end-to-end against a real search pipeline: setting smaller expansion
  values produces correspondingly shorter highlights in the formatted output,
  with the core reranked passage preserved.

### DIFF
--- a/src/tools/search/source-processing.test.ts
+++ b/src/tools/search/source-processing.test.ts
@@ -141,6 +141,63 @@ describe('expandHighlights content stripping', () => {
     expect(result.organic?.[0].content).toBeUndefined();
     expect(result.organic?.[0].references).toBeUndefined();
   });
+
+  test('honors a custom mainExpandBy when expanding highlights', () => {
+    const highlightText = 'KEYFACT';
+    // Boundary-free filler so expansion is governed purely by mainExpandBy,
+    // not by where natural separators happen to fall.
+    const content = `${'x'.repeat(1000)}${highlightText}${'y'.repeat(1000)}`;
+    const makeData = (): t.SearchResultData => ({
+      organic: [
+        {
+          ...makeOrganic('https://a.com'),
+          content,
+          highlights: [{ text: highlightText, score: 0.9 }],
+        },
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
thanks for the PR. note there are merge conflicts

PR #278, opened 2026-06-29
feat(run): add UserTraceContext to RunConfig for backend-agnostic trace attribution

dev-elie:
## Summary
Also see: https://github.com/danny-avila/LibreChat/issues/13529#issuecomment-4836097866

Adds `UserTraceContext { userId? }` to `RunConfig` so hosts can pass a human-readable user identifier to all configured trace backends (Langfuse, LangSmith, OTel, etc.) from a single place, rather than each backend re-deriving the same mapping.

When `RunConfig.user` is not set, the SDK falls back to resolving the userId from fields already present in `config.configurable` — so the improvement works without requiring a host upgrade.

## Breaking Change

Traces that previously showed a raw `user_id` string (e.g. a MongoDB ObjectId) from `configurable.user_id` will now show the user's email by default. Set `LIBRECHAT_TRACE_USER_ID_FIELD=id` to preserve the old behaviour.

## How to test

1. Configure a Langfuse (or LangSmith) backend and run an agent via LibreChat.
2. Open the Langfuse UI — the trace should show a human-readable `userId` (email by default) instead of a raw ObjectId.
3. Set `LIBRECHAT_TRACE_USER_ID_FIELD=name` (or `username` / `id`) and restart — verify the correct field appears.
4. Run unit tests: `npm test`

## Companion PR

danny-avila/LibreChat#14012 — wires `buildUserTraceContext(user)` into `createRun` and adds `tracing.userIdField` to the `librechat.yaml` config schema

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ coverage/**/*
 .cursor/
 .claude/
 .codex/
+.idea/
--- a/README.md
+++ b/README.md
@@ -115,6 +115,25 @@ const customHandlers = composeEventHandlers(
 );
 ```
 
+## Observability
+
+When Langfuse, LangSmith, or OTel tracing is configured, each run is attributed to the user that triggered it. By default the `userId` field in your trace backend is set to the user's **email address**.
+
+Set `LIBRECHAT_TRACE_USER_ID_FIELD` to change which field is used:
+
+| Value             | What appears as `userId`                                           |
+| ----------------- | ------------------------------------------------------------------ |
+| `email` (default) | `alice@example.com`                                                |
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Superseded by #514, which adds `userId` to `LangfuseConfig` and resolves it in one place for the stream, title, label, and phase handlers, keeping `configurable.user_id` as the fallback so hosts on the current default see no change. Host side lands in danny-avila/LibreChat#15667. Thanks for the companion work — closing this one.

PR #293, opened 2026-07-07
feat(hitl): declare multiSelect on AskUserQuestionRequest

danny-avila:
## Summary

Adds `multiSelect?: boolean` to the SDK's `AskUserQuestionRequest` interface (`src/types/hitl.ts`, source for `dist/types/types/hitl.d.ts`).

When `true`, the host UI may let the user pick several options; the resulting `AskUserQuestionResolution.answer` is the selected option `value`s joined by `", "`. Omitted/`false` → single-select. Only meaningful alongside `options`.

## Why

LibreChat's ask-user-question feature ([#14139](https://github.com/danny-avila/LibreChat/pull/14139)) passes a tool input containing `multiSelect` into `askUserQuestion()`. It works today only because the helper forwards the object verbatim into `interrupt()`, and librechat-data-provider's `Agents.AskUserQuestionRequest` mirror already declares `multiSelect?: boolean`.

Declaring the field in the SDK closes the type-contract gap so a future SDK change that validates or picks declared fields cannot silently drop it.

## Changes

- `src/types/hitl.ts`: add `multiSelect?: boolean` with doc comment; update `AskUserQuestionResolution.answer` JSDoc to cover the joined-values case.
- `src/tools/__tests__/hitl.test.ts`: new case asserting `multiSelect` rides through the `ask_user_question` interrupt payload and resumes with the `", "`-joined answer.

## Verification

- `jest` AskUserQuestion suite: 5 passed (incl. new test).
- `npm run build`: confirmed `multiSelect?: boolean;` lands in generated `dist/types/types/hitl.d.ts`.

## Related

Sibling to [#292](https://github.com/danny-avila/agents/pull/292) — both stem from the LibreChat #14139 ask_user_question rollout. That PR fixes the tool-error-completion contract across interrupt/resume; this one closes the type-contract gap. No file overlap; independent.

### DIFF
--- a/src/tools/__tests__/hitl.test.ts
+++ b/src/tools/__tests__/hitl.test.ts
@@ -3996,6 +3996,54 @@ describe('AskUserQuestion — interrupt + resume', () => {
     expect(resumedAnswer).toBe('production');
   });
 
+  it('carries multiSelect through the interrupt payload and resumes with the joined option values', async () => {
+    const { askUserQuestion } = await import('@/hitl');
+
+    let resumedAnswer: string | undefined;
+
+    const builder = new StateGraph(MessagesAnnotation)
+      .addNode('clarifier', () => {
+        const resolution = askUserQuestion({
+          question: 'Which environments?',
+          options: [
+            { label: 'Staging', value: 'staging' },
+            { label: 'Production', value: 'production' },
+          ],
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #296, opened 2026-07-10
🎚️ fix: Calibrate Context Pruning From Last-Call Usage, Not Accumulated Cache Reads

romy651:
## Summary

Fixes prune calibration when provider usage includes prompt-cache detail fields that are not directly comparable with the active context window.

The pruner now prefers the provider-reported `usageMetadata.input_tokens` for calibration and skips calibration for turns where the reported input is clearly not comparable, such as Bedrock cache-artifact turns with tiny `input_tokens` and very large cache creation/read detail fields.

## Root Cause

Cache detail fields can represent provider billing/cache accounting rather than the actual prompt window size. Folding those details into calibration could inflate the observed input and poison the running calibration ratio or resolved instruction overhead.

## Impact

This keeps pruning budget estimates stable when prompt caching is active, while preserving raw token maps and existing provider-specific cache semantics.

## Validation

- `npx tsc --noEmit`
- `npx eslint src/messages/prune.ts src/specs/token-accounting-pipeline.test.ts`
- `git diff --check HEAD^ HEAD`
- `npx jest src/specs/token-accounting-pipeline.test.ts`
- `npx jest src/specs/summarize-prune.test.ts src/specs/token-distribution-edge-case.test.ts src/specs/thinking-prune.test.ts`

### DIFF
--- a/src/messages/prune.ts
+++ b/src/messages/prune.ts
@@ -135,9 +135,10 @@ export type PruneMessagesParams = {
   usageMetadata?: Partial<UsageMetadata>;
   startType?: ReturnType<BaseMessage['getType']>;
   /**
-   * Usage from the most recent LLM call only (not accumulated).
-   * When provided, calibration uses this instead of usageMetadata
-   * to avoid inflated ratios from N×cacheRead accumulation.
+   * Fallback usage from the most recent LLM call only (not accumulated).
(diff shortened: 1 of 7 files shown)


danny-avila:
@codex review

PR #301, opened 2026-07-12
feat(openai): persist GPT-5.6 Responses state

usnavy13:
## Summary

Adds the OpenAI adapter support required for GPT-5.6 Responses reasoning continuity without native Programmatic Tool Calling.

- passes managed prompt-cache and safety fields through Chat Completions and Responses
- preserves `response.output` metadata for streaming and non-streaming Responses
- requests and replays encrypted reasoning for stateless/ZDR calls
- carries `reasoning.mode` and `reasoning.context`
- preserves cache-write usage metadata

## Scope

This is the base Agents change for the LibreChat GPT-5.6 Responses-state PR. Native PTC is intentionally excluded and is reviewed in a separate stacked PR.

## Verification

- focused managed-request and message-replay tests pass
- package build and TypeScript declaration build pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@langchain/google-vertexai": "2.2.0",
         "@langchain/langgraph": "^1.4.6",
         "@langchain/mistralai": "^1.2.0",
-        "@langchain/openai": "1.5.3",
+        "@langchain/openai": "1.5.5",
         "@langchain/textsplitters": "^1.0.1",
         "@langchain/xai": "^1.4.3",
         "@langfuse/langchain": "^5.4.1",
@@ -40,7 +40,7 @@
         "mathjs": "^15.2.0",
         "nanoid": "^3.3.7",
         "okapibm25": "^1.4.1",
-        "openai": "^6.35.0",
+        "openai": "^6.46.0",
         "uuid": "^11.1.1"
       },
       "devDependencies": {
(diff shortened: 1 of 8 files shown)


danny-avila:
Thanks a lot for this, @usnavy13 — genuinely useful work, and the managed-request + replay tests were the perfect spec.

I'm going to close this in favor of #303, which builds directly on your changes (tests reused, `Co-authored-by` you). The one structural difference: while reviewing I traced how our `LibreChatOpenAIResponses` sits on top of `@langchain/openai`, and it turns out upstream **already** implements the reasoning-continuity core — `convertMessagesToResponsesInput` does the `encrypted_content` replay + `response.output` reuse, and the Responses chat model already sends `previous_response_id`/`include`/`store` and passes the whole `reasoning` object through (so `reasoning.mode`/`reasoning.context` ride for free). Our class only overrides `_getReasoningParams` + `_getClientOptions`, so it inherits all of that.

That means the converter changes here re-implement behavior we already get from upstream, which would leave us re-syncing every future langchainjs Responses fix by hand. #303 keeps everything that's genuinely ours — managed prompt-cache breakpoints, `safety_identifier`, cache-write metadata, the encrypted-reasoning `include`, and the 1.5.5 bump — but injects those over upstream's converter output instead of forking it.

Not a knock on the approach at all — the inheritance path is easy to miss from outside the fork internals. Really appreciate you pushing this forward, and I'd love more contributions. 🙏

PR #311, opened 2026-07-15
fix subagent streamed result fallback

romy651:
## Summary

Fixes subagent executions returning `Task completed` or stale pre-tool assistant text when the child agent's final visible answer was only available from streamed `ON_MESSAGE_DELTA` events.

The executor now captures streamed message-delta text by message id, preserves message-id order, and uses the streamed result when the latest returned AI message has no visible text. Existing fallback behavior is preserved for non-streaming cases where only earlier AI text is available.

## Validation

- `npx jest src/tools/__tests__/SubagentExecutor.test.ts` passed: 81/81
- `npx eslint src/tools/subagent/SubagentExecutor.ts src/tools/__tests__/SubagentExecutor.test.ts` passed
- `git diff --check` passed

`npx tsc --noEmit` still fails on existing OpenAI response typing errors in `src/llm/openai/index.ts` around `prompt_cache_breakpoint`; those failures are unrelated to this change.

### DIFF
--- a/src/tools/__tests__/SubagentExecutor.test.ts
+++ b/src/tools/__tests__/SubagentExecutor.test.ts
@@ -374,6 +374,14 @@ describe('buildChildInputs', () => {
 describe('SubagentExecutor', () => {
   const config = makeConfig();
 
+  type ForwarderLike = {
+    handleCustomEvent?: (
+      eventName: string,
+      data: unknown,
+      runId: string
+    ) => Promise<void> | void;
+  };
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Thank you for identifying and working on this issue. I re-investigated it with live provider runs and opened #385 with a narrower canonical-message fix that avoids buffering forwarded stream events. Superseded by #385.

PR #313, opened 2026-07-16
fix(tools): stateful descriptions promise state the sandbox never keeps

danny-avila:
## Summary

The stateful `execute_code` / `bash_tool` descriptions promise the model in-memory state that the sandbox never keeps. Statefulness here is **filesystem-tier**: a session reuses one warm MicroVM (so `/mnt/data` persists), but every execution runs in a **fresh nsjail sandbox with its own PID + mount namespace** — when the call returns, that namespace is torn down and everything in it is reaped.

Measured against a live stateful backend, in one **warm** session (identical `/proc/sys/kernel/random/boot_id` across calls) with a filesystem control passing in the same run:

**`execute_code`** claimed *"variables, imports, and in-memory data from earlier successful calls are typically still available"*:
```
exec1: x = 42; import math      -> ok
exec2: print(x)                 -> NameError: name 'x' is not defined
exec3: print(math)              -> NameError: name 'math' is not defined
exec4: write /mnt/data/f.txt    -> ok      # control: same warm session
exec5: read  /mnt/data/f.txt    -> ok      # filesystem does persist
```
The shell PID is `2` on **every** call — a fresh PID namespace each time. There is no shared runtime; this isn't "occasionally resets," it never holds.

**`bash_tool`** claimed *"files (including /tmp) ... and running background processes from earlier calls typically persist"*:
- `/tmp` written in one call is **empty** in the next (private mount namespace)
- a backgrounded loop writing a heartbeat file **froze at its last tick** and never advanced (PID namespace reaped)

Both notes hedge with *"usually/typically ... may reset"*, which reads as rare degradation rather than a guarantee that never holds. A model trusting them writes `df = ...` in one call and `df.head()` in the next, or starts a server and assumes it's still listening — then hits an error the description told it to treat as unlikely.

- Rewrote `STATEFUL_ENV_NOTE` + `STATEFUL_CODE_PARAM_NOTE` (`CodeExecutor`) and `STATEFUL_BASH_NOTE` + `STATEFUL_BASH_PARAM_NOTE` (`BashExecutor`) to state what's true: the machine is warm and `/mnt/data` persists; each call is a new process, so variables, imports, shell state, cwd, `/tmp`, and background processes never carry over.
- Updated the JSDoc above each note to explain the architecture (warm VM + per-call jail) so the next editor doesn't reintroduce the claim.
- **Stateless variants are unchanged.**

## Change Type

## Testing

`bun test src/tools/__tests__/CodeExecutor.stateful.test.ts src/tools/__tests__/BashExecutor.test.ts` — **19 pass / 0 fail**. `tsc --noEmit` clean.

Note for reviewers: the existing tests **asserted the false wording** (`expect(d).toContain('usually share one runtime')`, `expect(stateful).toContain('typically still defined')`) — they encoded the bug, so they're updated as part of the fix. Added negative assertions (`.not.toContain(...)`) so the old claims can't silently return.

## Checklist

- [x] I have written tests demonstrating that my changes are effective

### DIFF
--- a/src/tools/BashExecutor.ts
+++ b/src/tools/BashExecutor.ts
@@ -58,17 +58,18 @@ Usage:
 `.trim();
 
 /**
- * Bash statefulness is filesystem-tier: on a warm session the machine (files
- * including /tmp, installed packages, background processes) persists between
- * calls, but each call may start a fresh shell — so shell variables and cwd
- * are NOT reliable, and the machine can be reset at any time. Only /mnt/data
- * is durable.
+ * Bash statefulness is filesystem-tier and scoped to `/mnt/data`. The machine
+ * is warm across calls, but each call runs in a fresh sandbox (new process
+ * tree + private /tmp), so background processes are reaped when the call ends
+ * and anything written outside /mnt/data is discarded. The note must not
+ * promise otherwise: a model told background processes survive will start a
+ * server in one call and assume it is listening in the next.
  */
 export const STATEFUL_BASH_NOTE =
-  'Session state (best-effort): commands in this conversation usually run on the same machine, so files (including /tmp), installed packages, and running background processes from earlier calls typically persist. Each call may still start a fresh shell — do not rely on shell variables or the working directory carrying over — and the machine may be reset at any time. Only /mnt/data is durable.';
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #314, opened 2026-07-16
fix: preserve current turns across compaction

AlexZahar:
## Summary

- persist semantic summary coverage through a source message ID; retain raw history if the latest summary coverage is missing, malformed, or unresolved
- preserve the latest structured user turn exactly through pruning and fail before provider invocation if it cannot fit
- inject mid-run summaries as neutral assistant history and use neutral default summarizer prompts
- support explicit, current-turn-only tool suppression from ToolMessage artifacts so bounded tools such as file_search can be disabled without affecting later turns

## Tool suppression contract

Tool results may return an artifact shaped as:

~~~json
{
  "toolControl": {
    "disableTools": ["file_search"]
  }
}
~~~

Only ToolMessage artifacts after the latest HumanMessage are considered. Parsing is structural and strict. Matching named tools are filtered while all other tools remain available. This is deliberately not a generic recursion finalizer.

## Why no recursion finalizer here

langgraph_step is absolute across checkpoint resumes, while recursionLimit is scoped to each invocation. A finalizer based on comparing those two values would prematurely finalize resumed runs. A correct generic finalizer needs an invocation baseline or dedicated graph state/node and is deferred from this patch.

## Verification

- 4 focused Jest suites: 251 passed, 1 skipped
- npx tsc -p tsconfig.json --noEmit
- npm run build
- git diff --check

## Release / consumer note

This PR intentionally does not bump the package version. Consumers such as Agentic Karl need a future @librechat/agents release newer than 3.2.65 containing this commit, or can pin this exact commit until that release is published.


AlexZahar:
@danny-avila please review

danny-avila:
@codex review

danny-avila:
Hi @AlexZahar, thanks for the work here. Could you please double-check and address the findings below where valid? I reviewed the current head, [`db3fa6886d`](https://github.com/danny-avila/agents/commit/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7), and reproduced these cases. I also double-checked the Codex findings, and they appear actionable.

### Compaction and summary handling

- **Allow summary-only continuation after full compaction.** With `retainRecent.turns: 0`, the summarize node can remove every message and leave a pending summary, but the [unconditional empty-message guard](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/graphs/Graph.ts#L2072) throws before `systemRunnable` can inject that summary. This makes a supported summary-only continuation fail on its next model call. This also confirms the [Codex finding](https://github.com/danny-avila/agents/pull/314#discussion_r3615764406).

- **Preserve legacy summary-only histories that have no coverage metadata.** The [new coverage validation](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/messages/format.ts#L1182) rejects older summary blocks that predate `coverage`. If the raw history was already compacted to only that summary, formatting removes the block and returns neither messages nor `initialSummary`, so the conversation cannot resume. This confirms the [Codex finding](https://github.com/danny-avila/agents/pull/314#discussion_r3615764411).

- **Correct token accounting for messages containing a summary block.** After the summary part is stripped from the assistant message, [the original full-message token count is still retained](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/messages/format.ts#L1558), while the summary token count is also added to instruction overhead. This counts the summary twice and can cause extra pruning or premature compaction. This confirms the [Codex finding](https://github.com/danny-avila/agents/pull/314#discussion_r3615764418).

- **Place the OpenRouter summary before the sole retained user message.** The [summary is appended as an AI message](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/agents/AgentContext.ts#L749), and the [single-message OpenRouter tail case](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/agents/AgentContext.ts#L785) produces `system -> current user -> assistant summary`. The summary therefore looks like an answer to the current request. The retained user message should remain last.

### Current-turn detection and tool suppression

- **Use the canonical real-user predicate when protecting the current request during pruning.** The [last `HumanMessage` lookup](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/graphs/Graph.ts#L1833) can select a hook-injected or skill/meta human message instead of the actual user request. I reproduced a case where an image-only user request was pruned while the injected hook message remained, so the preservation check passed even though the provider never received the real request.

- **Use the same real-user boundary for same-turn tool controls.** The [backward tool-control scan](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/graphs/Graph.ts#L127) stops at any `HumanMessage`. If a skill or hook injects a meta human message after a tool-disable artifact, the disable instruction is hidden and the tool stays enabled. This confirms the [Codex finding](https://github.com/danny-avila/agents/pull/314#discussion_r3615764420).

- **Resolve names for every supported tool schema shape.** [`getToolBindingName`](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/graphs/Graph.ts#L107) only reads a top-level `name`, so OpenAI-style `function.name` and Bedrock-style `toolSpec.name` schemas are not matched by suppression, even though those schema types are accepted as graph tools. This confirms the [Codex finding](https://github.com/danny-avila/agents/pull/314#discussion_r3615764414).

- **Carry the filtered tool set into fallback providers.** The primary path binds the [filtered resolved tools](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/graphs/Graph.ts#L1524), but the [fallback path receives the unfiltered `agentContext.tools`](https://github.com/danny-avila/agents/blob/db3fa6886db5726b117771fa7f6b3c5f9f2cc9c7/src/graphs/Graph.ts#L2277). A tool disabled for the current turn can therefore become available again after the primary provider fails.

Please add focused regression coverage for these edge cases, especially summary-only histories, the OpenRouter one-user-message path, injected meta human messages, schema-form tools, and fallback execution. Thank you!

danny-avila:
@codex review

danny-avila:
Superseded by #354, which lands the compaction-boundary fix from this PR against current `main`.

Thanks @AlexZahar — the problem you identified here is real and worth fixing. A summary's reach was inferred from where its content block sits in the payload, which cannot tell a retained recency tail from covered history, so the tail was silently dropped on the next run. #354 carries that fix: summaries now declare the first message compaction retained, anchored to the retained side rather than the covered one.

The other three changes here are not carried over. The `current_turn_exceeds_context` throw runs counter to #330, which made context overflow recoverable rather than fatal; and removing the `hasPendingCompactionSummary()` exception from the empty-messages guard breaks every `retainRecent.turns: 0` compaction — that is what the CI failures on this branch are (5 failures in `src/hooks/__tests__/compactHooks.test.ts`, reproduced locally).

The `toolControl.disableTools` contract is unrelated to compaction and would be welcome as its own PR. The three findings against it from the review above still apply: schema-shaped tools bypass the name filter, meta `HumanMessage`s break turn-scoping, and the tool-token budget is not recomputed after suppression.

Closing as superseded — no reflection on the work, just splitting it up so each piece can land on its own merits.

PR #317, opened 2026-07-18
fix: report provider model to Langfuse

AlexZahar:
## Summary
- prefer the provider-reported response model when exporting Langfuse generations
- normalize Azure Model Router values such as `model-routergpt-5.5-2026-04-24` to the concrete model/version
- preserve the existing Bedrock usage normalization

## Why
Langfuse currently receives the configured alias (`model-router`) through `model_name`, while Azure returns the selected model/version in `response_metadata.model`. The generic alias prevents Langfuse from matching model pricing and leaves generation cost details empty.

## Verification
- focused Langfuse callback suite: 10 tests passed
- ESLint
- TypeScript type-check
- package build

### DIFF
--- a/src/langfuse.ts
+++ b/src/langfuse.ts
@@ -92,14 +92,15 @@ function getLangfuseBedrockUsage(
   };
 }
 
-function cloneMessageWithUsage(
+function cloneMessageForLangfuse(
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #319, opened 2026-07-20
🛰️ fix: Drop Google Server-Side Tool Blocks on Cross-Provider Handoff

danny-avila:
## Summary

Switching a conversation to a Bedrock or Anthropic agent crashed with `Unsupported content block type: toolCall` when the history contained a Gemini turn that used a server-side tool (URL context, Google Search). Gemini emits those as `toolCall`/`toolResponse` parts carrying a `thoughtSignature`; the Google provider round-trips them, but the Bedrock and Anthropic converters had no branch for them and threw. Only Google can execute these blocks or validate their signatures, so I drop them on cross-provider handoff, mirroring the existing foreign-reasoning handling. The assistant's answer text is preserved.

- Added `FOREIGN_SERVER_TOOL_TYPES` (`toolCall`, `toolResponse`) to the Bedrock Converse converter and dropped these blocks on assistant turns, next to the existing `FOREIGN_REASONING_TYPES` drop; the empty-turn placeholder already covers a fully dropped turn.
- Dropped the same block types in the Anthropic `_formatContent`, gated to assistant turns via `isAIMessage`, with the existing empty-content placeholder as fallback.
- Kept the throw for genuinely unknown block types so real content is surfaced rather than silently discarded.
- Added `cross-provider-server-tools.test.ts` regression suites for both converters, with the repro block shaped like the production error (`URL_CONTEXT` args plus `thoughtSignature`), plus placeholder and still-throws cases.

## Change Type

## Testing

- `npm test -- cross-provider --testPathIgnorePatterns=worktrees`: all suites pass, including the new regression tests.
- `npm test -- src/llm/bedrock src/llm/anthropic --testPathIgnorePatterns=worktrees`: 20 suites, 355 tests pass.
- `npx eslint` on touched files and `npx tsc --noEmit` are clean.
- Manual repro path: run a Gemini agent with URL context on a conversation, then switch the conversation to a Bedrock or Anthropic agent and send a message. Previously this crashed the run; with this change the model responds normally.

## Checklist

### DIFF
--- a/src/llm/anthropic/utils/cross-provider-server-tools.test.ts
+++ b/src/llm/anthropic/utils/cross-provider-server-tools.test.ts
@@ -1,0 +1,110 @@
+import { AIMessage, HumanMessage } from '@langchain/core/messages';
+import type { BaseMessage } from '@langchain/core/messages';
+import { _convertMessagesToAnthropicPayload } from './message_inputs';
+
+/**
+ * Regression for cross-provider agent handoffs (Google → Anthropic): a Gemini
+ * turn that used a server-side tool (URL context, Google Search) leaves
+ * `toolCall`/`toolResponse` content blocks in history. The Anthropic converter
+ * has no branch for them and previously threw
+ * "Unsupported message content format", crashing the handoff. Only Google can
+ * execute these blocks or validate their thought signatures, so they are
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #323, opened 2026-07-21
fix: skip prefilled model turn for no-prefill Gemini models

danny-avila:
## Summary

Gemini **3.6 Flash** and **Gemini 3.5 Flash-Lite** newly reject (HTTP 400) any request whose `contents` array ends with a `model`-role turn:

> `Requests ending with a model turn are not supported unless the last part is a function response.`

This is **model-scoped**, not version-wide. Verified live against the Gemini API with identical prefilled requests:

| Model | Trailing `model` turn |
|---|---|
| `gemini-3.6-flash` | **400** |
| `gemini-3.5-flash-lite` | **400** |
| `gemini-3.5-flash` | 200 |
| `gemini-2.5-flash` | 200 |

A trailing model turn is only produced by prefill/continuation flows (e.g. a host editing an assistant reply and resubmitting it); normal chat and tool turns end with a user/function turn.

## Change

- `rejectsModelTurnPrefill(model)` — matches the no-prefill Gemini models (exact + versioned aliases), extensible as Google widens the restriction.
- `dropUnsupportedModelTurnPrefill(contents, model)` — for those models only, drops trailing `model` turns so the model generates fresh from the preceding user turn. **Strict no-op for every other model**, preserving working prefill behavior.
- Applied in `ChatGoogleGenerativeAI._generate` and `_streamResponseChunks` right before the request is built, where `this.model` is available.

## Tests

- 7 new unit tests in `src/llm/google/utils/common.test.ts` (predicate + strip behavior, including multi-turn, no-op cases, versioned aliases). Suite green.
- `tsc` clean; `npm run build` succeeds.
- The live integration suite (`llm.spec.ts`) only exercises `gemini-1.5/2.0/2.5/3-pro/3.5-flash` — none in the no-prefill list — so this change is a no-op for it.

Ref: https://ai.google.dev/gemini-api/docs/latest-model#api-changes-and-parameter-updates

Consumed by LibreChat's Gemini 3.6 Flash / 3.5 Flash-Lite support (danny-avila/LibreChat#14367).

### DIFF
--- a/src/llm/google/index.ts
+++ b/src/llm/google/index.ts
@@ -19,6 +19,7 @@ import type { GoogleClientOptions, GoogleThinkingConfig } from '@/types';
 import {
   convertResponseContentToChatGenerationChunk,
   convertBaseMessagesToContent,
+  dropUnsupportedModelTurnPrefill,
   mapGenerateContentResultToChatResult,
 } from './utils/common';
 
@@ -254,6 +255,7 @@ export class CustomChatGoogleGenerativeAI extends ChatGoogleGenerativeAI {
       this.client.systemInstruction = systemInstruction;
       actualPrompt = prompt.slice(1);
     }
+    actualPrompt = dropUnsupportedModelTurnPrefill(actualPrompt, this.model);
     const parameters = this.invocationParams(options);
     const request = {
       ...parameters,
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #343, opened 2026-07-27
🧪 fix: Transform ESM-Only Dependencies So Jest Can Load Them

danny-avila:
## The problem

`@mistralai/mistralai` is published `"type": "module"` with no CJS build at all, and `@langchain/mistralai` requires it from its own CJS output. Jest's CommonJS runtime cannot load that chain, so **any** suite that transitively imports `src/llm/providers.ts` died at collection:

```
SyntaxError: Unexpected token 'export'
  at Object.<anonymous> (node_modules/@mistralai/mistralai/esm/index.js)
  at Object.<anonymous> (src/llm/providers.ts:2:1)
  at Object.<anonymous> (src/llm/invoke.ts:16:1)
```

`providers.ts` imports `ChatMistralAI` purely to populate the provider map, so this hit tests with nothing to do with Mistral — every graph-level suite, every session suite, the hooks integration tests, and `src/llm/invoke.test.ts` itself.

## The fix

Exclude the two packages from `transformIgnorePatterns` and add a ts-jest transform for `.js`/`.mjs` with `allowJs`, so they are down-levelled to CJS on the way in.

`isolatedModules` and `diagnostics: false` on that entry are deliberate: these are third-party files, already type-checked upstream, and we want the syntax transform rather than a typecheck of `node_modules`.

## Measured

Using CI's own unit-shard invocation, with no credentials present so it matches the CI environment:

| | Suites failed | Suites passed | Tests failed | Tests passed |
|---|---|---|---|---|
| before | **44** | 129 | 30 | 2537 |
| after | **0** | 157 | **0** | 2911 |

374 tests that could never execute now run and pass. The credential-gated suites skip cleanly rather than hard-failing.

No measurable cost: **11.5s vs 12.5s** wall for the full unit run — a suite that fails to collect is not free either.

## Scope

`jest.config.mjs` only. No product code, no new dependencies (`ts-jest` is already the preset).

## Why now

This has been blocking real work. It is why #335 has no CI coverage of its core path and had to be verified with a live-credential script instead, and verifying one of its review findings required hand-rolling a throwaway jest config that stubs `@mistralai/mistralai` just to load a graph. Landing this first means the follow-on LibreChat PRs can have ordinary integration tests.

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -23,11 +23,22 @@ const config = {
   testEnvironment: 'node',
   testMatch: ['**/src/**/*.test.ts', '**/src/**/*.spec.ts'],
   transform: {
-    '^.+\\.tsx?$': ['ts-jest', { tsconfig: './tsconfig.json' }],
+    /**
+     * Declaring `transform` REPLACES the preset's entry, so the TypeScript
+     * rule has to be restated — including the `module: 'commonjs'` override
+     * the preset applies for us. The project targets `"module": "ESNext"`
+     * (tsconfig.json), and emitting that into Jest's CommonJS runtime fails
+     * at load with `ReferenceError: exports is not defined`.
+     *
+     * The object form merges over the project tsconfig rather than replacing
+     * it, so path aliases and the rest still apply.
+     */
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #347, opened 2026-07-28
🎯 feat: Tool Intent & Outcome Label Core

danny-avila:
## Summary

First slice of the **tool intent labels** feature (the fourth per-tool capability alongside `defer_loading`, `allowed_callers`, and `run_in_background`): a model-authored `intent` string declared as the **first property** of a tool's input schema, streamed by providers as the first key of the tool-call args so a host UI can render it as the call's live status label before the rest of the args exist — plus `outcome`/`outcome_patch` result fields through which a tool edits that sentence in place when the call settles.

This slice is machinery only. **No tool opts in yet** — schema injection for the native coding/search/subagent suite lands in a follow-up PR, and the LibreChat host capability (`tool_intents`) in a third.

### `src/tools/intentArg.ts` (new)

- `INTENT_ARG` / `INTENT_PROPERTY` / `INTENT_DESCRIPTION` — the frozen schema property, with the model instruction (write it FIRST; one sentence; present progressive; distinguish sibling calls).
- `withIntent(schema)` — non-mutating, idempotent first-key prepend (object key order is insertion order, and every provider serializer preserves it). Never added to `required`: the same schemas are callable from PTC, where no UI renders a label and forcing generated code to fabricate one is pure cost.
- `readIntent` / `stripIntent` — tolerant of both object and stringified-JSON args.
- `applyOutcome(intent, {outcome, outcome_patch})` — precedence: full replacement → first-occurrence case-sensitive span edit → mechanical leading-verb past-tense transform (`Searching…` → `Searched…`), unknown verbs pass through unchanged. Pure and dependency-free so hosts can import or mirror the exact resolution logic.

### Result threading

- `ToolExecuteResult` gains `outcome?` / `outcome_patch?`; `ProcessedToolCall` gains `outcome?`, emitted on `ON_RUN_STEP_COMPLETED` **only when the tool authored one** (bare-intent mechanical transforms are left to the host, so the wire never carries a label the host can derive itself).
- Covered paths: the event-driven batch loop, the early per-result fast path (`dispatchEarlyToolCompletion`), the eager streaming path (`dispatchEagerToolCompletions`), and the direct in-process path (artifact-borne fields via `readOutcomeFields`). Error results never settle a label — error/cancel framing stays a host concern.

### Tool output references

The top-level `intent` arg is excluded from `{{tool<i>turn<n>}}` placeholder substitution: it is a display label, not a data channel, and expanding a stored output there would dump up to the registry cap (~400 KB) into a single-line UI label and persist it with the message. Nested `intent` keys inside other args are unaffected.

## Testing

- `npx jest src/tools/__tests__/intentArg.test.ts` — 29 tests: first-key placement, non-mutation on frozen schemas, idempotence, required untouched, stringified-args tolerance, the full `applyOutcome` precedence table (case preservation, first-occurrence-only patches, unknown-verb passthrough, single-word intents), `resolveToolOutcome` / `readOutcomeFields` validation.
- `npx jest src/tools/__tests__/toolOutputReferences.test.ts` — includes the new root-`intent` exclusion test (top-level verbatim, nested still substituted).
- Focused runs of the suites touching edited files: `ToolNode.onResultCompletion`, `ToolNode.eagerEventExecution`, `ToolNode.outputReferences`, `ProgrammaticToolCalling`, `ToolSearch` — all pass (327 tests total).
- `npx tsc --noEmit` clean; `eslint` clean on all touched files (the one remaining warning in `ToolNode.ts` is pre-existing code outside this diff).

## Sequencing

1. **This PR** — core arg + outcome machinery.
2. Native tool schema injection (coding suite across all three engines, `web_search`, `subagent`, `skill`, `tool_search`) with a coverage-pin test.
3. LibreChat host capability `tool_intents` (injection for MCP/action tools, invoke-time strip, builder toggles) + UI streaming labels.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,6 +27,7 @@ export * from './tools/ReadFile';
 export * from './tools/skillCatalog';
 export * from './tools/ToolSearch';
 export * from './tools/ToolNode';
+export * from './tools/intentArg';
 export * from './tools/schema';
 export * from './tools/handlers';
 export * from './tools/local';
--- a/src/llm/openai/index.ts
+++ b/src/llm/openai/index.ts
@@ -1702,7 +1702,7 @@ class LibreChatOpenAIResponses extends OriginalChatOpenAIResponses {
         ]),
       ];
     }
-    return params;
+    return stripIntentFromStrictTools(params);
   }
 
   async completionWithRetry(
@@ -1952,7 +1952,7 @@ class LibreChatAzureOpenAIResponses extends OriginalAzureChatOpenAIResponses {
         ]),
       ];
     }
-    return params;
+    return stripIntentFromStrictTools(params);
   }
 
   async completionWithRetry(
--- a/src/llm/openai/index.ts
+++ b/src/llm/openai/index.ts
@@ -51,6 +51,7 @@ import {
   projectOpenAIResponsesToolMessageContent,
   projectToolStreamContentForProvider,
 } from '@/messages/core';
+import { INTENT_ARG, isIntentLabelProperty } from '@/tools/intentArg';
 import { isReasoningModel, _convertMessagesToOpenAIParams } from './utils';
 import { dropRepeatedScalarMetadata } from './streamMetadata';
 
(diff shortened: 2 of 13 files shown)


danny-avila:
@codex review

danny-avila:
Addressed all four findings in 1eb0e443:
- **Outcome bounding**: `resolveToolOutcome` collapses the label to a bounded single line (256 chars, whitespace-normalized) before emission.
- **Error labels**: authored `outcome`/`outcome_patch` now settle failed calls too; the status gate at all four completion sites is gone, so an unauthored error call stays unlabeled instead of risking a success-looking mechanical transform host-side.
- **Aggregation/persistence**: `ToolCallPart.outcome` added; the content aggregator's rebuilt tool call preserves it on final update and `SubagentExecutor.sanitizeProcessedToolCall` forwards it.
- **Effective args**: direct-path `PreToolUse.updatedInput` and HITL `edit` rewrites now record the effective args into `resolvedArgsByCallId` before `runTool`, so completion events (args + resolved labels) reflect what actually ran; placeholder substitution still overwrites with its post-substitution version when applicable.

@codex review

danny-avila:
Round-2 findings addressed in 1334684a:
- **Strict mode (P1)**: the (Azure)ChatOpenAI `invocationParams` overrides now drop the marker-identified optional `intent` from any tool auto-marked `strict: true` (the non-streaming `json_schema` structured-output path, which never renders a live label). Business `intent` params and explicitly-required labels are untouched; three new specs in `llm.spec.ts` pin it.
- **Error patches (P1)**: an error result's unmatched `outcome_patch` no longer falls through to the mechanical past-tense transform — failed calls label only with tool-authored text (`isError` threaded from all four completion sites).
- **Stringified args**: provider-style stringified JSON object args now preserve `{{tool…}}` placeholders inside the top-level `intent` (parse → key-aware transform → re-serialize); other fields still substitute, plain strings keep the raw fast path.
- **Literal patches**: `outcome_patch.to` applies via a replacement callback so `$&`-style tokens stay verbatim.
- **Event contract**: `ToolEndEvent.tool_call` now declares `output`/`progress`/`outcome`.

@codex review

danny-avila:
Round-3 findings addressed in 5e8306b7:

- **Responses strict tools (P1)**: both Responses delegates now run the sanitizer — round 2 only wrapped the Completions pair. Worth recording that the *trigger* differs per API, which the finding's framing didn't capture: Completions infers `strict` from a `json_schema` response_format, while the Responses API flattens tools to `{type, name, parameters, strict}` and leaves `strict: null` for that same request — an explicit `strict` (per-call option or on the tool) is what turns it on there. Both mechanisms are covered and tested; my first draft test asserted the Completions trigger on the Responses path and correctly failed.
- **Business `intent` + references**: fixed. `withIntent` deliberately declines to inject when a tool already declares `intent`, so the resolver can't treat the key as UI-only — `resolve()` now takes `substituteIntentKey`, and `ToolNode` sets it per call from the tool's own schema (marker-guarded), including the snapshot views and the stringified-args path. Unresolved refs inside a business `intent` are reported again too.
- **Returned error ToolMessages**: fixed, and it was a real terminal-event gap rather than only a label gap — `undispatchedToolErrors` is populated exclusively in the `catch` path, so a tool returning `status: 'error'` had the handler credited for a dispatch that never happened and got no `ON_RUN_STEP_COMPLETED` at all. Ownership is now tracked explicitly (`handlerOwnedToolErrors`, set when the handler dispatches *or throws*, preserving the existing throw-means-dispatched reasoning) instead of inferred from the absence of a marker. The regression test fails against the old guard and passes against the new one — verified both directions.

@codex review

danny-avila:
Round-5 findings addressed in fcd438b8 — both confirmed real:

- **Error ownership is now batch-local.** Moved off the instance into per-invocation state threaded through `RunToolBatchContext`. The interleaving that exposes it: invocation 1 batches a throwing call (handler claims ownership) alongside a slow call that keeps its output loop pending; while parked, invocation 2 reuses the id and RETURNS an error message and consumes the other invocation's marker. I wrote the regression test for exactly that shape and verified it emits an **empty** completion list against the old instance-scoped code versus the expected id against the new — so this dropped a call's only terminal event, not just its label. Also closes the stale-marker leak when an interrupt aborts a batch before its output loop reads it.

  I moved the pre-existing `undispatched` set along with it rather than leaving one half instance-scoped: the two are set and consumed together, and that set's consume-on-use was itself a workaround for this exact scoping problem.

- **Mixed direct+event pre-resolution** now passes `substituteIntentKey`. `dispatchToolEvents` consumes those pre-resolved args without re-resolving, so this was the last unguarded path — a business `intent` parameter there kept the literal `{{tool…}}` and wasn't even reported unresolved.

Full tools + search suites green (1065 passing). #349 rebased on top (8587bb08).

@codex review

rio100:
cool! had AI translate it for me lol

**In everyday terms:**

Imagine you have a toolbox where all the tools are labeled with what they do ("hammer: for nails", "saw: for cutting wood"). This update makes sure every tool gets a label, and now when you use a tool, it can tell you what it accomplished in a consistent way.

**What you might notice as a non-developer:**

    If you use this software, your agents might work more reliably
    The reporting of what agents did might be clearer and more consistent
    It's setting the foundation for future improvements to the tool system

PR #349, opened 2026-07-28
🎯 feat: Intent-First Schemas for Native Tools

danny-avila:
## Summary

Second slice of tool intent labels, **stacked on #347** (retarget to `main` once that merges). Opts the native tool suite into the `intent` arg: every coding tool on every engine, plus `web_search`, `subagent`, `skill`, and `tool_search`, now declares `intent` as the **first schema property** — optional, never in `required`.

### Injection strategy: schemas, not factories

Cloudflare has no file tools of its own — `CloudflareSandboxTools.ts` imports the local factories and reuses the shared bash/code schemas — so injecting at the schema definition covers all three engines from a handful of edit sites:

| Site | Covers |
|---|---|
| 6 local coding schemas (`local/LocalCodingTools.ts`) + `local/CompileCheckTool.ts`, via `withIntent(...)` | local **and** cloudflare-sandbox engines |
| `BashExecutionToolSchema` / `CodeExecutionToolSchema`, inline first key | remote, local, and cloudflare `bash_tool` / `execute_code` (incl. the stateful `build*Schema` variants, which spread the base) |
| both PTC schema factories (+ `intent` on the closed `ProgrammaticToolCallingJsonSchema` type) | `run_tools_with_code` / `run_tools_with_bash` on all engines — the local/cloudflare variants inherit through their property spreads |
| `ReadFileToolSchema`, `SkillToolSchema`, `SubagentToolSchema` (+ the runtime `buildSubagentToolParams` variant), `ToolSearchToolSchema` (+ `createToolSearchSchema`), `WebSearchToolSchema` | the rest of the native surface |

### The two real hazards this had to defuse

- **`...rest` wire-body leaks:** `BashExecutor`/`CodeExecutor` spread unrecognized args straight into the Code API POST body. `intent` is now destructured out exactly like `runtime_session_hint`. Every other native body reads specific fields, so the optional extra key is inert.
- **Frozen-instance validation crash:** LangChain's JSON-schema validator (`@cfworker/json-schema`) stamps `__absolute_uri__` onto every subschema it dereferences, which throws on a frozen object. The property is now always embedded as a **copy** of the frozen canonical `INTENT_PROPERTY` (fix backported to #347's `withIntent`). Caught by the execution suites, now pinned by a unit test.

### Outcome authoring

- `outcomeFieldsFromResult`: completion events honor **artifact-borne** `outcome`/`outcome_patch` on the event-driven paths too, so a `content_and_artifact` tool authors its settled label the same way everywhere.
- `web_search` is the first author: `Found N results for "query"` rides its artifact next to the citation data.

### Coverage pin

`intentCoverage.test.ts` asserts every name in `LOCAL_CODING_BUNDLE_NAMES` and `CLOUDFLARE_CODING_TOOL_NAMES` — plus `read_file`, `skill`, `subagent` (static + runtime), `tool_search`, `web_search`, the stateful schema variants, and both PTC runners — emits `intent` first and never requires it. "Did we get them all" is now a failing test instead of a review question.

## Testing

- 447 tests across the 11 affected suites (`intentArg`, `intentCoverage`, `localToolNames`, `LocalExecutionTools`, `CloudflareSandboxExecution`, `BashExecutor`, `ProgrammaticToolCalling`, `SkillTool`, `SubagentTool`, `ToolSearch`, `ReadFile`) — all pass.
- `npx tsc --noEmit` clean; eslint clean on the diff (remaining warnings in touched files pre-date it).

### DIFF
--- a/src/stream.ts
+++ b/src/stream.ts
@@ -41,7 +41,7 @@ import {
 } from '@/utils/truncation';
 import { TOOL_OUTPUT_REF_PATTERN } from '@/tools/toolOutputReferences';
 import { safeDispatchCustomEvent } from '@/utils/events';
-import { resolveToolOutcome } from '@/tools/intentArg';
+import { resolveToolOutcome, outcomeFieldsFromResult } from '@/tools/intentArg';
 import { isGoogleLike } from '@/utils/llm';
 import { getMessageId } from '@/messages';
 
@@ -860,9 +860,11 @@ async function dispatchEagerToolCompletions(args: {
         maxToolResultChars
       ).content;
     }
-    const outcome = resolveToolOutcome(record.request.args, result, {
-      isError: result.status === 'error',
-    });
+    const outcome = resolveToolOutcome(
+      record.request.args,
+      outcomeFieldsFromResult(result),
+      { isError: result.status === 'error' }
+    );
 
     try {
       const dispatched = await safeDispatchCustomEvent(
--- a/src/stream.ts
+++ b/src/stream.ts
@@ -39,9 +39,9 @@ import {
   calculateMaxToolResultChars,
   truncateToolResultContent,
 } from '@/utils/truncation';
+import { resolveToolOutcome, outcomeFieldsFromResult } from '@/tools/intentArg';
 import { TOOL_OUTPUT_REF_PATTERN } from '@/tools/toolOutputReferences';
 import { safeDispatchCustomEvent } from '@/utils/events';
-import { resolveToolOutcome, outcomeFieldsFromResult } from '@/tools/intentArg';
 import { isGoogleLike } from '@/utils/llm';
 import { getMessageId } from '@/messages';
 
--- a/src/tools/BashExecutor.ts
+++ b/src/tools/BashExecutor.ts
@@ -3,7 +3,6 @@ import fetch, { RequestInit } from 'node-fetch';
 import { HttpsProxyAgent } from 'https-proxy-agent';
 import { tool, DynamicStructuredTool } from '@langchain/core/tools';
 import type * as t from '@/types';
-import { INTENT_PROPERTY } from '@/tools/intentArg';
 import {
   BASH_SHELL_GUIDANCE,
   CODE_ARTIFACT_PATH_GUIDANCE,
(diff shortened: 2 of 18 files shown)


danny-avila:
@codex review

danny-avila:
Addressed in 51586d64: the web_search outcome now counts the actual result collections (organic, top stories, news, images, videos, places) instead of `references`, which only tracks highlight-embedded links. Branch also rebased on #347's review fixes.

@codex review

danny-avila:
Round-2 finding addressed in 61ebf34e: `createSearchTool` builds its runtime schema dynamically (provider-dependent `country`), so production instances still led with `query` while the coverage test checked the unused static def. The dynamic schema now leads with `intent`, and the coverage pin exercises the factory itself (serper + searxng variants). Rebased on #347's round-2 fixes.

@codex review

danny-avila:
Round-3 findings addressed in b6ef994a: dropped the narrating JSDoc (the key order and coverage pin already encode the invariant), and ran the repo's sorter over the seven files this PR touches. Left the unrelated pre-existing sort failures elsewhere in the tree alone rather than sweeping them into this diff. Rebased on #347's round-3 fixes.

@codex review

danny-avila:
Round-4 findings addressed in 45f41111 — both were real and the first is the more serious of the two:

- **Caught search errors now author a failure label.** The processor reports provider/processing failures through `data.error` while still returning NORMALLY, so the `ToolMessage` carries success status; a bare intent would have settled mechanically from "Searching…" to "Searched…" and shown a failed search as a successful one. That case now emits `Search failed for "<query>"`, which takes precedence over any partial results.
- **Count covers every rendered kind.** Added the `peopleAlsoAsk` collection and the `knowledgeGraph`/`answerBox` singletons, so a response containing only those is no longer labeled as zero results (previously it got no authored outcome at all).

I extracted the computation as the pure exported `resolveSearchOutcome` instead of leaving it inline in the tool body — it's now directly testable, and `outcome.test.ts` covers error precedence over partial results, empty-string errors, all six collection kinds, both singletons, and the genuine zero-result passthrough that deliberately falls back to the host's mechanical transform.

@codex review

PR #350, opened 2026-07-28
🧹 fix: Apply Every Delta Content Part; Remove Dead SplitStreamHandler

danny-avila:
## Summary

Two changes from the LibreChat reasoning-splitting investigation (see danny-avila/LibreChat#14494 for context):

### 1. Content aggregator: apply every delta content part

`createContentAggregator` read only `content[0]` of `ON_MESSAGE_DELTA` / `ON_REASONING_DELTA` payloads via `getFirstContentPart`, silently dropping the remaining entries of any multi-part delta. That's reachable today: Google server-side tool chunks dispatch several reasoning parts in a single event (`dispatchGoogleServerSideToolStreamContent`), and the SSE emitters already iterate all parts — so streamed output and saved content could diverge. Every entry now reaches the step's slot in order, preserving the existing concatenation semantics; empty arrays still no-op.

### 2. Remove the dead SplitStreamHandler

Dead since LibreChat stopped consuming it in danny-avila/LibreChat#10533 (Dec 2025); nothing imports it here either. Removed:

- `src/splitStream.ts` (the `blockThreshold: 4500` legacy splitter) and its package-index export
- The `SplitStreamHandlers` / `CustomChunk*` types that existed only for it
- `src/mockStream.ts` — its only consumer was the splitter's test file

**Breaking note:** `SplitStreamHandler`, `SEPARATORS`, `SplitStreamHandlers`, and `CustomChunk*` leave the public API surface. LibreChat has zero references; external consumers relying on the removed splitter would need to vendor it.

### Test moves + new coverage

`splitStream.test.ts` was misnamed — most of it was ContentAggregator coverage (empty deltas, resumed tool completions, physical content indices, provider-specific parts). Those suites move **intact** to `src/aggregator.test.ts`, joined by new multi-entry delta tests (multi-text, multi-think, mixed cadence). Only the splitter-specific suites are deleted with the splitter.

## Validation

- `tsc --noEmit` clean
- `jest src/aggregator.test.ts src/stream.test.ts src/graphs/__tests__/Graph.reasoning.test.ts`: 54/54 passing
- Full jest run: **no new failures** — the 14 pre-existing failures (provider live specs needing API keys, langfuse-routing's `--experimental-vm-modules` issue) are byte-identical on a clean checkout, verified via stash A/B


danny-avila:
@codex review

PR #353, opened 2026-07-29
🏷️ refactor: Intent Label Refinements - No Tense Map, Terser Description, Shared Marker

danny-avila:
Follow-up to #347 / #349, informed by the first real-provider run of the feature.

## Drop the tense verb map

`applyOutcome` no longer rewrites a leading gerund to past tense. A closed English verb list is wrong three ways at once:

- **It never fires for non-English labels.** The design expects the model to answer in the user's language, so intents arrive as "Buscando…", "検索中…". A 22-entry English map is permanently dead for those users — the settled label already behaved differently depending on locale.
- **It fires for some siblings and not others.** A live run produced *"Recording the location of the OAuth callback router"*; "Recording" was not in the map. Put that beside a mapped "Searched…" in the same group and one card reads past tense, the other present, for no reason a user can perceive. Never transforming is more coherent than transforming sometimes.
- **It enumerates a vocabulary** in a feature whose premise is a free-form sentence.

Completion is better carried by UI state — the shimmer stopping, the icon settling — which is language-neutral and always consistent. Tools that want past tense still say so explicitly via `outcome` / `outcome_patch`, which remain the deliberate, unbounded path.

## Trim the description

502 → 289 chars (~126 → ~72 tokens). It rides every opted-in schema on every request, so this returns roughly **800 tokens per request** across a full coding bundle. What is kept is load-bearing: first-position placement (the entire streaming mechanism), the one-sentence present-progressive form, who reads it, and the sibling rule. What went is redundant — "Do not exceed one sentence" restated "one sentence".

## Export `INTENT_LABEL_MARKER`

The description's opening words double as the discriminator separating the injected label from a tool's own business parameter named `intent`. Host applications reimplement the same strip/sanitize passes and duplicated it as a string literal; if the copies drift the host silently stops recognizing SDK-native labels and **fails open** — labels stay in schemas, opt-outs stop working, no error. One definition now, imported downstream.

## Add `withoutIntent()`

Native schemas apply `withIntent` at module scope, so an embedder that renders no status label had **no lever at all** — they paid the tokens unconditionally. This is the opt-out. Marker-guarded, so a tool's own `intent` parameter is never removed.

## Testing

`src/tools`: 1139 passing. `intentArg.test.ts` covers the no-rewrite contract across English, non-English, lowercase and single-word labels; marker/description invariants (opens with the marker, stays ≤300 chars); and `withoutIntent` including the business-parameter guard and a `withIntent` round-trip. Typecheck and lint clean, imports sorted.

Verified live before writing this: real Anthropic, two sibling calls to one MCP tool, distinct first-position labels.

### DIFF
--- a/src/tools/__tests__/intentArg.test.ts
+++ b/src/tools/__tests__/intentArg.test.ts
@@ -282,7 +282,7 @@ describe('resolveToolOutcome', () => {
     ).toBe('Search failed for OAuth handling');
   });
 
-  it('never falls back to the mechanical transform on a failed call', () => {
+  it('never reuses the in-flight intent as a failed call\'s settled label', () => {
     expect(
       resolveToolOutcome(
         args,
--- a/src/tools/__tests__/intentArg.test.ts
+++ b/src/tools/__tests__/intentArg.test.ts
@@ -13,6 +13,7 @@ import {
   readOutcomeFields,
   resolveToolOutcome,
 } from '../intentArg';
+import { ReadFileToolSchema } from '../ReadFile';
 
 describe('withIntent', () => {
   const base: JsonSchemaType = {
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
Addressed in f893c0e4 — the finding was correct and the stale contract was the more dangerous half of the change: an embedder implementing against that JSDoc would have rebuilt the exact locale-dependent inconsistency this branch removes.

`resolveToolOutcome` now documents the real contract: a bare intent is displayed unchanged, hosts must NOT rewrite it (with a pointer to `applyOutcome` for why the tense transform is deliberately absent), and an unmatched patch on a failed call resolves to `undefined` rather than reusing the in-flight intent. Swept the rest of the module too — the only surviving mentions of the transform are the two explaining why it isn't there.

@codex review

danny-avila:
Fixed in 4ba65b09 — and the finding was right about severity. `ProcessedToolCall.outcome` is the *exported* contract embedders implement against, so a stale instruction there does more damage than the internal one: a host following it faithfully rebuilds the locale split and the inconsistent sibling labels this branch exists to remove. It now states the actual rule — display the intent unchanged, never rewrite its tense — with the reasoning inline so it does not get "restored" later.

My earlier sweep was too narrow (I grepped `intentArg.ts` and `types/tools.ts` only), so a repo-wide pass turned up two more: `resolveSearchOutcome`'s JSDoc described a zero-result search as deferring to "the host's mechanical past-tense transform", and two test names still described the removed behavior. All corrected. A full-tree grep now returns only the passages that explain the transform's deliberate absence.

@codex review

danny-avila:
Both fixed in d002cfa9 — and together they meant the advertised opt-out did not actually work on its primary target, which is worse than either finding alone.

- **Readonly schemas**: correct, and the irony is the point — the SDK's native schemas are `as const`, so `withoutIntent(ReadFileToolSchema)` failed TS2345 on the very schemas the helper exists for, forcing a cast to use the documented API. The parameter is now `IntentStrippableSchema`, widening `required` to `readonly string[]`. The new test calls `withoutIntent(ReadFileToolSchema)` directly, so it is a compile-time assertion rather than a runtime one.
- **`required` pruning**: also correct, and the strict-mode path makes it reachable rather than theoretical — OpenAI strict function schemas require every property in `required`, so a normalized schema hits exactly this. Leaving a dangling entry yields invalid JSON Schema and the provider rejects the request instead of the embedder quietly opting out. `required` is now filtered, and dropped entirely when `intent` was its only entry.

`src/tools`: 1142 passing, typecheck and lint clean.

@codex review

PR #356, opened 2026-07-29
🧵 feat: Previous-Headers Continuity for Activity Labels

danny-avila:
## Summary

Threads committed headers from earlier batches into the activity-label prompt, so consecutive same-activity batches extend the run's story instead of restating a line the user is already reading.

Each label call currently sees only its own batch, its reasoning excerpts, and the assistant's last text — it has no idea what the previous header said. In a real 9-batch run that produced two pairs of near-duplicate headers out of nine lines: a setup batch and its payoff batch each announced the same finding, and two consecutive network probes both reported "blocked". Read as the commit log the header register is aiming for, ~44% of the lines repeated their neighbour.

## Changes

- **`types/activityLabel.ts`** — `previousLabels?: string[]` on `RunActivityLabelOptions`. Run order, most recent last. Documented as COMMITTED labels only: a pending slot's text is empty, and a dropped fill never surfaced to the user, so neither belongs in context that shapes the next header.
- **`prompts/activityLabel.ts`** — renders the list as the prompt's FIRST section, capped at `MAX_PREVIOUS_LABELS` (3). Dropped wholesale under any active tool-output redaction policy, alongside reasoning excerpts and intent: per-agent Langfuse overlays mean an earlier header may have been generated under ANOTHER agent's weaker policy, and a handoff must not leak that phrasing into this trace.
- **`run.ts`** — `generateActivityLabel` passes the field through to the builder.

Purely additive; absent the field the prompt is byte-identical to today's.

## Evidence

Measured with an eval corpus that replays captured production payloads verbatim plus synthetic cases for the modes the real run never hit (all-failed, partial, parallel batches, rapid near-duplicate batches, entry overflow). 17 cases / 28 steps, `claude-haiku-4-5`, 3 samples per variant, sequences chaining each generated label into the next step's context.

Continuity context:

- **eliminated restatement** — zero restate incidents vs. the blind baseline's recurring ones;
- **produced correct setup→payoff pairs** on consecutive same-activity batches: `Wrote marker file to /tmp/persist-probe.txt` → `Confirmed marker file persists across tool calls`, where the blind variant wrote the same sentence twice;
- **prevented premature-conclusion labels** — an unexpected find. Without it, the model labelled the *setup* batch with the conclusion (`Confirmed X persists` on the write step, before anything had proved it). That is a correctness bug in the label, not a style preference;
- differentiated repeated probes by leading with what was new rather than re-announcing the shared finding.

Cost: +12% per label call from the larger prompt (a full 9-label run measured $0.0048 before).

## Consumer

LibreChat threads committed labels at request-build time — not claim time, since a batch claims its slot while the previous batch's fill is still in flight, which is exactly the rapid-consecutive case where continuity matters most. Host-side accumulator and resume seeding landed in danny-avila/LibreChat#14391; this field activates the SDK-traced path. Hosts on an older SDK pass the field harmlessly and see it rendered by their own fallback prompt.

## Test plan

- `src/specs/activity-label-prompt.test.ts` — 3 new cases: section renders first in run order with the oldest dropped at the cap; omitted when empty or absent; dropped under an active redaction policy.
- Full file green (10 tests), `tsc --noEmit` clean, eslint clean (the one warning at `activityLabel.ts:55` is pre-existing on `main`).

### DIFF
--- a/src/prompts/activityLabel.ts
+++ b/src/prompts/activityLabel.ts
@@ -33,6 +33,23 @@ export function truncateForLabel(value: string, maxLength: number): string {
   return value.slice(0, Math.max(0, maxLength - 1)) + '…';
 }
 
+/**
+ * Reduces a committed label to bounded single-line data.
+ *
+ * Sections in this prompt are delimited by blank lines, so a label carrying
+ * embedded newlines could otherwise forge an apparent `Tool calls:` or
+ * `Label:` section. Unlike every other input here, previous labels re-enter
+ * the prompt on EVERY later batch, so one malformed result — plain model
+ * noncompliance, or injection surfacing through a tool result — would
+ * persistently steer unrelated later labels rather than affecting one. The
+ * clip bounds the same way `lastAssistantText` and reasoning excerpts are
+ * bounded: oversized headers must not inflate later requests past the fast
+ * model's window and starve the run of labels entirely.
+ */
+function sanitizePreviousLabel(label: string): string {
+  return truncateForLabel(label.replace(/\s+/g, ' ').trim(), PREVIOUS_LABEL_LIMIT);
+}
+
 const ABORT_SERIALIZATION = Symbol('abort-label-serialization');
 
 /**
@@ -77,6 +94,10 @@ function serializeForLabel(value: unknown, limit: number): string {
 const INPUT_CONTEXT_LIMIT = 200;
 const MAX_THINKING_EXCERPTS = 4;
 const MAX_PREVIOUS_LABELS = 3;
+/** Per-label bound. A header is 5-9 words; anything past this is
+ *  noncompliance or payload, and previous labels are the one input that
+ *  RE-ENTERS the prompt on every later batch of the run. */
+const PREVIOUS_LABEL_LIMIT = 200;
 /** A label is 5-9 words; no batch needs more than this many entries to
  *  produce one, and the cap keeps a 200-call programmatic batch from
  *  building an enormous prompt out of per-field-bounded pieces. */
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
Round-1 findings addressed in `baf8d5fe`. Both share one root cause I had underweighted: previous labels are the only input that **re-enters the prompt on every later batch**, so one malformed label persistently steers unrelated later labels instead of affecting a single request.

- **Section forgery (P2)**: `sanitizePreviousLabel` collapses whitespace before interpolation, so an embedded newline can no longer mint an apparent `Tool calls:`/`Label:` section out of a noncompliant result or injection surfacing through a tool output. Fixed at the source too — `extractLabel` in `run.ts` now collapses whitespace, so `generateActivityLabel` cannot return a multi-line header for a host to feed back as continuity context.
- **Per-label bounding (P2)**: the same helper clips each label at `PREVIOUS_LABEL_LIMIT` (200), matching how `lastAssistantText` and reasoning excerpts are already bounded. Array-length capping alone did not bound prompt size, so a few oversized headers could exceed the fast model's window and starve the rest of the run of labels.
- **Empty-after-sanitize**: labels that reduce to nothing are filtered and the section is omitted entirely when none survive, so no bare bullet implies a missing header.

Tests: a multi-line label yields exactly one header bullet with the injected framing inert inside it (and exactly one real `Tool calls:` and one `Label:` section survive the render); a 5,000-char label is clipped rather than inlined; an all-blank list omits the section. 15 tests green, `tsc --noEmit` clean.

Also worth recording for reviewers: the `MAX_PREVIOUS_LABELS` cap of 3 was challenged and is now measured rather than assumed. An unbounded-history variant scored no better on the eval corpus (restatement was already fully eliminated at 3 — it is inherently a recency problem), while prompt growth is linear and unbounded: +82 input tokens by batch 9, extrapolating to roughly +250 on a ~550-token prompt at the `activityMaxPerRun` default of 20.

@codex review

PR #360, opened 2026-07-30
🧪 chore: SDK-Side Activity-Label Prose Eval Harness

danny-avila:
## Summary

Ports LibreChat's activity-label prose eval harness ([LibreChat #14527](https://github.com/danny-avila/LibreChat/pull/14527)) into the SDK, with one structural difference that is the point of the port: **the user prompt is rendered by the real `buildActivityLabelPrompt` from `src/prompts/activityLabel.ts`**, not a hand-port. A change to the builder (sections, truncation, framing) is measured here as it will ship.

`corpus.cjs` / `report.cjs` / `captured.json` are byte-identical to the LibreChat originals so results compare across repos; `checks.cjs` carries one deliberate divergence from review (namespaced MCP tool-name normalization in echo matching — backport candidate for the LibreChat copy). Fidelity of the two builders was verified before porting: 57 renders (every synthetic payload step × three previous-label depths), 0 byte mismatches.

## What's in the harness

- **`run.ts`** — runner on the repo's existing `tsconfig-paths-bootstrap.mjs` script convention. Captured verbatim steps get their continuity section rendered through the real builder (empty-batch render, terminal sliced off), so production's label sanitation and 3-label cap apply exactly rather than being reimplemented. Framing hypotheses are applied as marker-exact substitutions that degrade to the control when a marker doesn't match, never corrupting a sample.
- **`variants.ts`** —
  - `sdk-default`: `ACTIVITY_LABEL_PROMPT`, what a host gets when it passes no `prompt`. Never measured before this.
  - `host-shipped`: pinned fixture of LibreChat's `ACTIVITY_INSTRUCTION` (dev@a07c0e4ae8), the string every production label call actually receives.
  - `guard-heading` / `guard-full`: the host fallback builder's "What it called, and what came back (do not restate these):" framing (+ `Header:` terminal), which has never been on the live SDK path.
- `rescore.cjs` — re-grade stored results after a metric change, no re-spend.
- npm scripts `label:eval` / `label:rescore`; `results/` gitignored.

```bash
npm run label:eval -- --samples 3
npm run label:eval -- --dry --cases mega-batch   # render only, no API calls
npm run label:rescore
```

Needs `ANTHROPIC_API_KEY` (env or repo `.env`). ~$0.04/variant per 3-sample pass on claude-haiku-4-5.

## Sweep results (3 samples × 28 steps × 4 variants, re-run after the review fixes)

| variant | len | opener | tool-echo | template | top opener |
|---|---:|---:|---:|---:|---|
| sdk-default | 4 | 5 | 0 | 2 | Checked ×14 |
| host-shipped | 16 | 0 | 1 | 6 | Found ×21 |
| guard-heading | 10 | 0 | 1 | 6 | Confirmed ×23 |
| guard-full | 11 | 0 | 1 | 4 | Confirmed ×27 |

The review's start-of-prompt framing fix matters here: the original sweep silently rendered the control framing on six synthetic steps under the guard variants, so this table supersedes it. Pooled across three independent 3-sample runs (two here, one LibreChat-side), `guard-full` halves template-redundancy flags versus control (9 vs 18) and every guard arm has fewer length violations; per-run variance is real, so read the per-case tables before acting on any single sweep.

Two findings:

1. **The guard framing holds up directionally across three independent runs** (pooled: template flags 18 control vs 9 guard-full, fewer length violations in every guard arm, no per-case regressions by eye). Porting the guard heading into the builder is a reasonable follow-up PR, gradeable by this harness in-repo.
2. **`sdk-default` is format-tight but content-weak**: fewest length violations and ~20% cheaper input, but attempt-oriented labels ("Tested network access", "Checked latest versions of Node.js, Deno, and Bun" — the answer never becomes the line) and the run's only generic-opener violations. Quantifies the quality cliff for hosts that don't pass their own instruction.

The aggregate table is a regression guard; the per-case tables in `results/latest.md`, read by eye, are the real instrument — the production failure this corpus was built around (informationally redundant, lexically varied headers) scores below the automated overlap threshold.

## Post-rebase note (after #361/#362/#363 merged)

Rebased onto `main`. The `guard-heading`/`guard-full` hypothesis variants **shipped into the builder via #363**, so they are pruned here per the follow-up noted in that PR: the registered variants are now `sdk-default` and `host-shipped`, both rendering the shipped framing through the real builder, and `applyFraming`'s current-builder markers moved to the new heading/terminal (captured verbatim prompts predate the change and degrade to control by design). The sweep tables above are the pre-ship measurements that justified #363.


danny-avila:
@codex review

danny-avila:
Addressed the review in `cea56f42`:

- **Transport errors bypass retry** — fixed: fetch rejections are caught inside the attempt loop and retried like 429/500/529; after three attempts they become a recorded `error` record instead of rejecting the pool and discarding the run.
- **Single-line label normalization** — fixed: the harness now applies the production extractor's exact normalization (`.replace(/\s+/g, ' ').trim()` + quote strip), so graded/chained labels match what users receive and newlines can't break the per-case tables.
- **`Record<string, unknown>`** — fixed: explicit `DryRunRecord | ErrorRecord | LabelRecord` union, plus typed signatures at the CJS module boundaries.
- **Latency across retries** — fixed: `latencyMs` spans the full attempt sequence including backoff.
- **Import order** — fixed to the documented package / type / local convention.
- **Continuity terminal assumption** — fixed: the terminal is derived from an empty builder render, so the extraction survives the very `Header:` rename `guard-full` evaluates.
- **`--samples` validation** — fixed: `--samples` and `--concurrency` must be positive finite integers.
- **Namespaced tool-name echo matching** — fixed in `checks.cjs`: separator runs collapse to single spaces and the post-namespace tail is matched as a phrase. Single middle segments (e.g. `github`) are deliberately excluded — they're legitimate label subjects, and flagging them would over-report. This is now the one deliberate divergence from the LibreChat copy (noted in both file headers); backport candidate for LibreChat's `scripts/activity-labels/checks.js`.
- **Variant-specific word bounds** — declined: a single shared `len` rule is what keeps columns comparable across variants and repos; `sdk-default`'s 5-word minimum vs the shared 4-word floor is visible in the `avg words` column, and splitting the rule per variant would quietly decouple the two harnesses' metrics.

@codex review

danny-avila:
Round 2 addressed in `9df2ad1f` — all four findings were real, and two were sharp catches:

- **NUL delimiter in `rescore.cjs`** — confirmed: the chain key contained literal NUL bytes (inherited from the LibreChat original — my editor rendered them invisibly), and `git diff --stat` was indeed classifying the file as binary. Rewritten as `\0` escapes: identical runtime key, textual diffs restored. Flagged for the LibreChat backport as well.
- **Start-of-prompt heading** — confirmed empirically (`--dry --cases bare-batch` rendered the control framing under `guard-heading`): a prompt with no previous labels, intent, or excerpts starts directly with `Tool calls:\n`, which the boundary marker missed. `applyFraming` now handles the start-of-prompt case. This diluted the guard variants on six synthetic steps, so I re-ran the full sweep and updated the PR body's results table — the guard's direction holds (pooled across three runs: template flags 18 control vs 9 guard-full), with honest per-run variance noted.
- **Body reads outside the retry try** — fixed: `response.json()`/`response.text()` now sit inside the attempt's try, so a connection reset after headers retries instead of rejecting the pool.
- **Partially-unknown `--variants`/`--cases`** — fixed: any unresolved name fails up front listing the available names, instead of completing a billed run without its control column.

@codex review

danny-avila:
Round 3 addressed in `8aed30c2` — all five accepted:

- **Unknown CLI flags** — `parseArgs` now rejects them; a mistyped `--dryy` can no longer fall through to a fully billed default sweep.
- **Rescore corpus drift** — the runner persists a sha256 corpus fingerprint in the results JSON; `rescore.cjs` hard-refuses a stored run whose corpus has since drifted, and warns (but proceeds) on pre-fingerprint files.
- **Variant-ordered queue bias** — the task queue is now sample → case → variant, so adjacent queue slots cycle through the arms instead of correlating variant identity with elapsed run time and rate-limit pressure.
- **Numeric sample sort** and **keyed record index** in `report.cjs` — both fixed; both flagged as backport candidates alongside the checks/rescore divergences.

@codex review

danny-avila:
Round 4 addressed in `56ada309` — all four accepted:

- **Duplicate selections** — `resolveSelection` now rejects repeated `--variants`/`--cases` names (a repeat scheduled the same paid arm twice and merged both runs under one aggregate row while the keyed report showed only the last).
- **Spelled-out counts** — `COUNT_ECHO` now catches "three tool calls" as well as digits; noted as the second deliberate `checks.cjs` divergence pending backport.
- **Rescore in a fresh checkout** — `results/` is created before writing `latest.md`, so rescoring an archived JSON by explicit path no longer throws `ENOENT` after all the work is done.
- **Quadratic queue dispatch** — the pool consumes an index into the immutable task array instead of `shift()`.

@codex review

danny-avila:
Round 5 addressed in the latest commit — both accepted:

- **Auth failures are now run-fatal** — a 401/403 throws through the pool on the first response (verified live with an invalid key: one request, immediate `ERR authentication failed`), instead of issuing hundreds of identical doomed requests and writing an all-error report.
- **Opener tallies are case-folded** — `Found`/`found` count as one opener, so a sentence-case violation can no longer masquerade as register diversity. Third `report.cjs` divergence noted for the backport.

@codex review

danny-avila:
Round 6 addressed in the latest commit — both accepted:

- **Fatal classification before the body read** — the 401/403/404 check now precedes `response.text()`, and the detail read is `.catch(() => '')`-guarded, so a dropped error-body stream can no longer demote a definitive failure to a retryable transport error.
- **Unknown `--model` is run-fatal** — 404 joins the fatal set. Verified live: an invalid key and a bogus model each kill the run on the first response.

@codex review

danny-avila:
Round 7 addressed in the latest commit — both accepted:

- **Honest `AggregateRow` contract** — the asserted `Record<string, string | number>` was false (rows carry a nested `flagCounts` object and pre-formatted string fields); replaced with the explicit row shape `aggregate()` actually returns, so the reporting boundary is genuinely type-checked.
- **Option values validated** — value-taking flags reject a missing value or a following `--flag`; `--model --dry` no longer eats dry-run mode, and a trailing `--model` fails up front instead of 400-ing the entire sweep per task. Verified live in both forms.

@codex review

danny-avila:
Round 8 addressed in the latest commit (note: that round reviewed the pre-rebase head `4b72822b`; the branch has since been rebased onto main and aligned with the shipped #363 framing — all three findings still applied and are fixed on the rebased head):

- **Empty labels mirror production** — a 200 whose normalized label is empty becomes an error record ("production would commit no header") instead of a len:0 success, matching `extractLabel`'s `{}` return.
- **Prototype-safe opener tallies** — `firstWords` is a null-prototype object; `Constructor` now tallies as `constructor ×1` (verified) instead of colliding with `Object.prototype.constructor`, and `__proto__` no longer vanishes.
- **Dry-run continuity** — dry mode seeds the chain with captured production labels (or a marked placeholder), so multi-step prompts render the `Previous headers` section for inspection without a live spend.

@codex review

danny-avila:
Round 9 addressed in the latest commit — all three accepted:

- **Single-line error diagnostics** — HTTP bodies, fatal details, and transport messages collapse whitespace before storage, so a multiline gateway response can no longer split its per-case table row.
- **Dry runs persist everything** — all rendered prompts land in `results/dry-latest.md` (56/56 verified, including the later multi-step prompts that demonstrate the three-label continuity cap); the console still previews three.
- **Billed empty responses keep their usage** — the empty-label error record carries its tokens and `aggregate()` counts them, so an all-empty variant reports its real cost instead of $0.

@codex review

danny-avila:
Round 10 addressed in the latest commit — all three accepted:

- **Per-attempt deadline** — each API attempt carries `AbortSignal.timeout(60s)`; a hung connection surfaces as a retryable transport failure instead of pinning a worker and blocking the report forever.
- **Latency retained on error records** — `aggregate()` includes it (verified: a 900ms success + 1200ms billed-empty error report mean 1050ms), so failure-heavy arms show real wall time.
- **Aggregate opener cell escapes pipes** — matching the per-case renderer (verified with a `a|b` opener).

@codex review

danny-avila:
Round 11 addressed in the latest commit — accepted: 408/502/503/504 join the retryable set alongside 429/500/529, so a completed transient gateway failure retries instead of biasing an arm as a recorded error.

@codex review

PR #361, opened 2026-07-30
🧭 fix: Detach Langfuse Roots From Foreign Ambient Spans + Trace-Shaping Docs

danny-avila:
## Summary

Live production traces on Langfuse cloud surfaced three trace-shaping gaps. Investigation traced all three to one root cause, plus one independent naming leak. This PR fixes them, documents the trace-shaping invariants in AGENTS.md, and was verified against real Langfuse cloud ingestion.

### Root cause: foreign ambient OTEL spans hijack trace identity

LibreChat executes runs inside its host instrumentation's OpenTelemetry spans (HTTP server spans on the global provider). `@langfuse/langchain`'s `CallbackHandler` parents **root observations** on `context.active()`, so every trace root inherited a foreign span that is never exported to Langfuse. Observed live (trace `305e2aa9…`, `e7f635…`):

- **Null trace input/output**: roots arrived with dangling parents, so `shapeRootSpan` (which requires a true root) never fired — trace input/output stayed `null` and the root kept type `SPAN` instead of `agent`.
- **Merged agent + title traces**: an agent run and the previous turn's title generation fired inside one request context inherited the *same* foreign trace id — one merged trace with racing names (`AgentRun` vs `TitleRun`) and unioned tags (`agent` + `title`). Exactly the "title-tagged traces named AgentRun" seen in production.
- **`deterministicTraceId` silently broken**: the seeded id generator only runs for true roots, so an inherited ambient trace id bypassed it entirely.

### Fixes

1. **Foreign ambient span detachment** (`langfuse.ts`, `langfuseSpanRegistry.ts`, `instrumentation.ts`): `RoutingLangfuseSpanProcessor` registers every span it exports in a WeakSet; `ScopedLangfuseCallbackHandler` strips non-registered (foreign) ambient spans from the OTEL context for *detached runs* — roots, or starts whose `parentRunId` the handler never tracked (the base handler's runMap-miss fallback would otherwise parent them on the ambient span; this is how a detached subagent's model spans leaked). Langfuse-managed ambient spans still parent roots, so hosts can group runs under their own observations deliberately.
2. **Title runtime scope seed** (`run.ts`): `generateTitle` now seeds its runtime scope with `title-<runId>`, mirroring the activity-label path's documented fix — an inherited parent seed can no longer collapse the title into the run's trace.
3. **Ephemeral agent-id span rename** (`langfuseTraceShaping.ts`): the outer workflow node is named with the bare agent id; ids matching LibreChat's `encodeEphemeralAgentId` format (`endpoint__model___sender[____index]`) now reduce to their stable sender name — `bedrock__claude-sonnet-5___ClickHouse Agent` → `ClickHouse Agent` — so switching models no longer renames the span. Persisted `agent_*` ids (already stable) are untouched.

Plus the **AGENTS.md “Langfuse Trace Shaping” section**: module map, the invariants (from Langfuse-team feedback in #288/#316 + the new self-contained-identity rule), verification steps, and pointers to Langfuse's agent-facing docs (skill, docs MCP, llms.txt).

## Verification

- `npx jest langfuse deterministic-trace-id` — new specs: export-time ephemeral-id renames (5 cases) and two routing integration tests (foreign ambient span detached / managed ambient span preserved).
- Full CI-equivalent suite: **3360 passed, 0 failed**.
- **Live probe against Langfuse cloud** (env `ci`): ran an `AgentRun` stream + `TitleRun` title inside a simulated foreign host span with `deterministicTraceId: true`. Result: two separate traces at their seeded ids — agent trace root typed `agent` with trace input `"What is the probe answer?"` / output `"Deterministic probe answer."`, span named `Probe Agent`; title trace root typed `chain`, tags `[librechat, probe, title]`; nothing on the foreign trace id.

## Out of scope (pre-existing, filed separately)

`langfuse-routing.integration.test.ts` has two failures that reproduce on `main` with a clean `npm ci` and are invisible in CI (validate.yml ignores `integration.test.ts`): when two tenants' runs execute concurrently, some spans are exported through the *other* tenant's Langfuse processor (cross-tenant leak, e.g. the whole tenant-b summarization tree on tenant-a's credentials). Root cause is ambient-scope-first config resolution crossing concurrent runs' ALS contexts — needs its own precedence design; flagged as a follow-up task.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,17 +4,18 @@
 
 `@librechat/agents` is a TypeScript library for LLM agent orchestration — tool calling, multi-agent graphs, message formatting, streaming, and provider abstraction (Anthropic, Bedrock, VertexAI, OpenAI, Google). Published as `@librechat/agents` on npm. This is a major backend dependency of [LibreChat](../LibreChat/CLAUDE.md) (same team).
 
-| Path            | Purpose                                             |
-| --------------- | --------------------------------------------------- |
-| `src/messages/` | Message formatting, caching, content processing     |
-| `src/graphs/`   | LangGraph-based agent graphs (single + multi-agent) |
-| `src/llm/`      | Provider-specific LLM wrappers and utilities        |
-| `src/tools/`    | Tool definitions and search                         |
-| `src/agents/`   | Agent definitions and handoff logic                 |
-| `src/types/`    | Shared TypeScript types                             |
-| `src/common/`   | Enums, constants                                    |
-| `src/run.ts`    | Main run orchestration                              |
-| `src/stream.ts` | Streaming logic                                     |
+| Path                                         | Purpose                                                                              |
+| -------------------------------------------- | ------------------------------------------------------------------------------------ |
+| `src/messages/`                              | Message formatting, caching, content processing                                      |
+| `src/graphs/`                                | LangGraph-based agent graphs (single + multi-agent)                                  |
+| `src/llm/`                                   | Provider-specific LLM wrappers and utilities                                         |
+| `src/tools/`                                 | Tool definitions and search                                                          |
+| `src/agents/`                                | Agent definitions and handoff logic                                                  |
+| `src/types/`                                 | Shared TypeScript types                                                              |
+| `src/common/`                                | Enums, constants                                                                     |
+| `src/run.ts`                                 | Main run orchestration                                                               |
+| `src/stream.ts`                              | Streaming logic                                                                      |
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

danny-avila:
@codex review — addressed both findings in 60c24eae (destination-keyed span registry + strict encoded-id validation with root gating)

danny-avila:
@codex review — round-2 feedback addressed in 1653ceed: rename gated on matching langgraph_node observation metadata (verified against production span attributes), decoded nodes typed as agent observations, and toolOutputTracing moved out of destination identity into the processor cache key.

PR #362, opened 2026-07-30
🔭 fix: Keep Generation-Root Observation Input in Langfuse Traces

danny-avila:
## Problem

`shapeRootSpan` (from #288/#316, per Langfuse-team feedback) reduces every **root** span's observation input/output to the last user question and assistant answer. For chain/agent roots that is pure presentation — the child generation still records the complete prompt. But a generation that **is** the trace root — a bare `model.invoke` with no wrapping chain, which today means the **activity-label** path — has no child. The reduction discarded the `SystemMessage` (the entire label instruction, the quality surface of the feature) from the only place it is traced, making the instruction impossible to A/B against production traffic.

Diagnosed empirically with an OTLP-capture probe against the published 3.3.8 dist: the span is created with the full `[system, user]` message array as its input, and exports with only the user prompt — the rewrite happens in `shapeLangfuseSpan` at export time. Titles are unaffected (they pipe through a chain, so their root is a chain span); the main stream is unaffected (its root is a LangGraph chain span).

## Fix

In `shapeRootSpan`, skip the **observation-level** input/output reduction when the root is itself a generation span. **Trace-level** input/output still reduce, so the trace list keeps showing the question and answer exactly as before — including the "trace input begins with `Previous headers in this run…`" property that the continuity rollout validation relies on.

## Verification

- Two new tests: a generation root (the exact attribute shape `@langfuse/langchain` exports for the label path) keeps its full observation input/output while trace input reduces; a non-generation root keeps the existing full reduction. All 19 trace-shaping tests pass.
- End-to-end OTLP capture against this branch's built dist: the exported label generation now carries both messages (`system prompt present: true`), span still exports as the shaped `llm` root, trace input still reduces to the user prompt.

### DIFF
--- a/src/langfuseTraceShaping.ts
+++ b/src/langfuseTraceShaping.ts
@@ -254,11 +254,20 @@ function shapeRootSpan(span: MutableSpan): void {
     parseAttributeValue(span.attributes[outputKey]),
     'assistant'
   );
-  if (question != null) {
-    span.attributes[inputKey] = question;
-  }
-  if (answer != null) {
-    span.attributes[outputKey] = answer;
+  /** A generation that IS the trace root — a bare `model.invoke` with no
+   *  wrapping chain, i.e. the activity-label path — is also the only
+   *  observation carrying its own prompt: reducing its observation input
+   *  would discard the SystemMessage from the one place it is traced.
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #363, opened 2026-07-30
🏷️ feat: Reference-Material Framing for the Activity-Label Prompt

danny-avila:
## Summary

Ports the host fallback builder's entries framing onto the live SDK path: the entries heading becomes `What it called, and what came back (do not restate these):` and the terminal cue becomes `Header:`.

LibreChat's fallback prompt builder has always framed the tool list this way — its `runtime.ts` documents that without the guard "the model tends to read the list as the thing to summarize and hands back a transcription of it" — but that guard had never been on the live path, which sends a bare `Tool calls:` … `Label:`. The divergence between the two builders was the defect; this closes it in the measured direction.

## Measurement (the point of #360)

Measured **before** porting, via the eval harness (#360, ported from LibreChat #14527):

- Across three independent 3-sample sweeps (28 corpus steps each; two SDK-side, one LibreChat-side), the guard framing produced fewer template-redundancy flags than the bare framing (pooled 9 vs 18) and fewer length violations in every arm, with no per-case regressions by eye.
- A post-change sweep against **this builder** (local merge with the harness branch) confirms the as-shipped rendering: the production-instruction arm scored its best length-violation count of any run (8, vs 12–16 for prior bare-framing controls) with zero tool-name echoes.

## Also in this change

- The section-forging defense test (`activity-label-prompt.test.ts`) now attacks the new markers — a malicious previous label attempting to inject the new heading/terminal still flattens to inert data.
- The `sanitizePreviousLabel` rationale comment follows the rename.

## Interaction with #360

Once this merges, the harness's `guard-heading` / `guard-full` hypothesis variants become no-ops by design (their marker substitutions no longer match, so they degrade to the new baseline); pruning them is a one-line follow-up on the harness side. Merge order between the two PRs doesn't matter.

### DIFF
--- a/src/prompts/activityLabel.ts
+++ b/src/prompts/activityLabel.ts
@@ -37,8 +37,8 @@ export function truncateForLabel(value: string, maxLength: number): string {
  * Reduces a committed label to bounded single-line data.
  *
  * Sections in this prompt are delimited by blank lines, so a label carrying
- * embedded newlines could otherwise forge an apparent `Tool calls:` or
- * `Label:` section. Unlike every other input here, previous labels re-enter
+ * embedded newlines could otherwise forge an apparent entries section or
+ * `Header:` cue. Unlike every other input here, previous labels re-enter
  * the prompt on EVERY later batch, so one malformed result — plain model
  * noncompliance, or injection surfacing through a tool result — would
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #364, opened 2026-07-30
🔒 fix: Stamp Langfuse Runtime Scopes With Run Identity

danny-avila:
## Summary

Fixes a **cross-tenant Langfuse span-routing leak**: with two runs executing concurrently in one process, spans from one tenant's run were exported through the *other* tenant's `LangfuseSpanProcessor` — tenant data leakage in multi-tenant deployments. Reproduced deterministically by the two `langfuse-routing.integration.test.ts` tests that have been failing on `main` (invisible in CI, which ignores `integration.test.ts`).

## Root cause (verified empirically)

LangChain executes non-awaited handler callbacks on a **process-wide background queue** — `consumeCallback` in `@langchain/core` pushes onto a module-singleton `p-queue` with `concurrency: 1` whenever `awaitHandlers` is false (the default unless `LANGCHAIN_CALLBACKS_BACKGROUND=false`). Tasks queued by run B while the queue drains run A's tasks execute in **run A's async continuation**, carrying A's AsyncLocalStorage and OTEL context.

Instrumented runs showed tenant-B's handler callbacks firing with `als=pk-tenant-a otel=pk-tenant-a` — the handler's ambient-scope-first resolution (`resolveLangfuseConfigForSpan(ctx) ?? this.langfuse`, intended for same-run agent overlays) then adopted A's config, and `RoutingLangfuseSpanProcessor` routed B's spans to A's processor. The ambient context at callback time is fundamentally untrustworthy for tenant identity.

## Fix

Runtime scopes and handlers now carry the **identity of the run they belong to** (`runId` on `LangfuseRuntimeContext`, the OTEL scope context, and handler creation params). In `withRuntimeContext`:

- ambient scope stamped with **the handler's own run** → adopted (same-run agent overlays keep winning, pinned by the existing overlay spec);
- ambient scope stamped with a **different run** → foreign: the handler's own configuration and seed apply;
- unstamped scope + unstamped handler → existing scope-first semantics (host-managed handlers unchanged).

Stamped call sites: stream (`this.id`), title (`title-<id>`), and activity-label (`activity-label-<id>-<seq>`) paths in `Run`; model-invoke handler, agent-overlay, and provider-fallback scopes in `Graph` (`this.runId`). Detached subagent graphs stamp their child run id, so subagent spans keep their current parent-trace placement through the child's own re-stamped scope.

## Verification

- `npx jest langfuse-routing`: **7/7** — including the two tests failing on `main` since before #361 (cross-tenant parallel flows and parallel summarization).
- Full Langfuse suite `npx jest langfuse deterministic-trace-id`: **137/137** (first fully green run), including the agent-overlay spec this fix must not break.
- Full CI-equivalent suite: **3370 passed, 0 failed**.
- New unit regressions in the CI-covered `langfuse-callbacks` spec (CI never runs the integration specs): foreign-scope rejection and same-run overlay adoption.

## Notes

- Builds on #361's detached-run handling; `detachForeignAmbientSpan` now receives the run-correct destination key.
- AGENTS.md "Langfuse Trace Shaping" invariants gained a "Runtime scopes are run-stamped" entry.
- Worth considering separately: un-ignoring `integration.test.ts` in validate.yml now that the suite is green.


danny-avila:
@codex review

danny-avila:
@codex review — all three P1s addressed in 01bda4f1: generated stamps for runId-less graphs, own tool-output policy on foreign scopes, and `<run>#<agentId>` sub-scopes verified against callback langgraph metadata for parallel fan-out agents.

danny-avila:
@codex review — round-2 findings addressed in 04468f45: foreign-scope rejection now replaces (clears) the scope instead of merging, and agent overlay identity is a separate scope field rather than a '#'-encoded stamp.

danny-avila:
@codex review — round-3 findings addressed in 8b7df3bd: ToolNode scopes stamp their executing agent, and foreign-scope rejection now clears @langfuse/tracing's propagated identity attributes (and the foreign active span) before re-propagating the handler's own identity.

danny-avila:
@codex review — round-4 P2 addressed in e0d064da: callback-node comparison checks a verbatim agent-id match before decoding internal node prefixes.

danny-avila:
@codex review — round-5 findings addressed in 7a46faf1: handlers restore their run's aggregate tool-output policy and trace name on rejection, explicit agentId metadata beats node-name parsing (ToolNode stamps it too), and derived scope ids carry opaque per-execution components.

danny-avila:
@codex review — round-6 findings addressed in 941fd157: stream scopes share the graph's opaque per-execution stamp (public ids still drive deterministic seeds), and the explicit agent lookup accepts the summarizer's agent_id key.

danny-avila:
@codex review — round-7 P2 addressed in dc375bff: the explicit agent lookup prefers the summarizer-owned agent_id over inherited agentId.

danny-avila:
@codex review — round-8 P2 addressed in 0e7a69eb: all identity-stamping components (including the summarizer) now overwrite the canonical agentId at their own invoke, so spread order — not key priority — decides the closest stamper.

danny-avila:
@codex review — round-9 P2 addressed in 51f6243d: the model path stamps the canonical agentId verbatim, outside createLangfuseTraceMetadata's length filtering.

danny-avila:
Closing out the automated review cycle here — nine rounds of findings addressed (rounds 1–3 were real cross-tenant/policy leaks, all fixed with regressions). The two round-10 P2s are acknowledged but out of the library's usage contract, so they are not being changed in this PR:

1. **Per-invocation stamps for reused graphs**: a single `StandardGraph` instance is not safe for concurrent invocations regardless of Langfuse — it holds mutable per-run state (the state reducer writes `this.messages`/`startIndex`; content data and checkpoint scopes are instance-level). One graph instance = one execution is the supported model (`Run.create` and `SubagentExecutor` both build fresh graphs), and the per-instance stamp matches that contract.

2. **Empty-string agent ids**: an agent id of `""` breaks agent-context keying and run-step attribution across the library, not just scope trust; the `!== ''` guard deliberately treats empty as "no identity." Meaningful, non-empty agent ids are a basic input invariant.

If either constraint should instead be hardened (e.g. validating non-empty agent ids at graph construction), happy to do that as a follow-up.

PR #369, opened 2026-07-31
🩹 fix: Keep Truncated Tool-Call Inputs JSON Objects on Anthropic Replay

danny-avila:
## Problem

The `validate / summarization tests (anthropic)` job intermittently fails (~2/15 main runs, e.g. [d8ab3494's run](https://github.com/danny-avila/agents/actions/runs/30559695284)) with:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.3.content.0.tool_use.input: Input should be an object"}}
```

## Root cause

**Not a model-side `max_tokens` truncation.** The #280 `OutputTruncationError` guard works — verified live end-to-end: a genuinely truncated tool-call turn throws and never enters state.

The real chain, reproduced live by intercepting `createStreamWithRetry` payloads in a descending-budget loop matching the CI test:

1. In a summarization-enabled run under a tight budget, the pruner's fit-to-budget pass calls `preFlightTruncateToolCallInputs` with an effective budget of a few dozen tokens → per-input cap ≈ **20 chars**.
2. A streamed `tool_use` block keeps its input as the raw partial-JSON **string**; it exceeds the cap, and the `{_truncated, _originalChars}` envelope (~38 chars) *also* exceeds it — so `createBoundedTruncationValue` returned **`null`** for both the block's inline `input` and its `tool_calls[].args` mirror.
3. The projection **mutates the live state array**, so the nulls persist into summarization-retained history.
4. On the next Anthropic call, the outbound sanitizer only handled *string* inputs (`typeof null === 'object'`), so the request shipped `"input": null` → 400.

The flakiness comes from an asymmetry: the escaped string serializes ~7 chars longer than the args object, so short inputs (`9 * 9`) kept their args and self-healed via the `tool_calls` restore path, while longer ones nulled both and failed.

## Fix (both ends)

- **`src/messages/prune.ts`** — `createBoundedTruncationValue` returns `{}` instead of `null` when even the empty envelope overflows the cap, so graph state never carries a non-object input. Also upgrades `serializeToolCallInput`'s degenerate output from `'null'` to `'{}'` on the OpenAI arguments path.
- **`src/llm/anthropic/utils/message_inputs.ts`** — new `coerceAnthropicToolUseInput` guarantees every outbound `tool_use`/`server_tool_use` input is a JSON object, applied after all restore paths in `_formatContent` and in `_convertLangChainToolCallToAnthropic`: complete-JSON strings parse; null / arrays / strings parsing to non-objects degrade to `{}`.

## Tests

- `src/specs/prune.test.ts` — regression: a cap below the truncation envelope yields `{}` (never `null`) for both the inline block input and `tool_calls[].args`; updated the `serializeToolCallInput(undefined, 4)` expectation from `'null'` to `'{}'`.
- `src/llm/anthropic/utils/streaming-tool-input.test.ts` — regression suite for non-object input replay: null input + null args ships `{}`; null input + intact args restores the args; string inputs parsing to non-objects (`'123'`, `'[1,2]'`, `'"text"'`, `'null'`) coerce to `{}`; complete JSON-object strings still parse; the string-content `tool_calls` path coerces null args.
- Live: the `cross-provider summarization` spec passes with the fix, and the payload-intercepting repro loop — with summarization firing on the previously-failing turn — produces zero bad payloads.

## Out of scope

The co-occurring `final_context_overflow` failures (e.g. runs 30446609385, 30389041546) are a **distinct** tight-budget mode — instructions + summary consume the entire descending test budget and the pre-invoke guard correctly refuses — confirmed to fire independently of the 400. Tracked separately.

### DIFF
--- a/src/llm/anthropic/utils/message_inputs.ts
+++ b/src/llm/anthropic/utils/message_inputs.ts
@@ -388,6 +388,34 @@ function _ensureMessageContents(
   return updatedMsgs as (SystemMessage | HumanMessage | AIMessage)[];
 }
 
+/**
+ * Anthropic requires `tool_use.input` to be a JSON object; anything else is
+ * rejected with a 400 (`tool_use.input: Input should be an object`).
+ * History can carry non-object inputs: streaming leaves the raw partial-JSON
+ * string on the content block, and upstream context projections can leave
+ * `null` (observed live on a summarization-retained tool call replayed after
+ * compaction). A string is parsed when it forms complete JSON; every other
+ * shape degrades to `{}` — the call already executed, so the replayed input
+ * is informational and an empty object is the safe representation.
+ */
+export function coerceAnthropicToolUseInput(
+  input: unknown
+): Record<string, unknown> {
+  let candidate: unknown = input;
+  if (typeof candidate === 'string') {
+    try {
+      candidate = JSON.parse(candidate);
+    } catch {
+      return {};
+    }
+  }
+  return typeof candidate === 'object' &&
+    candidate !== null &&
+    !Array.isArray(candidate)
+    ? (candidate as Record<string, unknown>)
+    : {};
+}
+
 export function _convertLangChainToolCallToAnthropic(
   toolCall: ToolCall
 ): AnthropicToolResponse {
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #370, opened 2026-07-31
🩹 fix: Coerce Non-Object toolUse Inputs in the Bedrock Converse Formatter

danny-avila:
## Problem

Bedrock twin of the Anthropic replay hardening in #369. Two related holes in `src/llm/bedrock/utils/message_inputs.ts`:

1. **Persisted nulls ship as-is.** Conversations persisted before #369 can carry tool calls whose inline `input` and `tool_calls[].args` were truncated to `null` by the pruner's old envelope-overflow branch. Both `convertAIMessageToConverseMessage` (materialize-from-`tool_calls` path) and `convertFromV1ToChatBedrockConverseMessage` (v1 `tool_call` block + `tool_calls` fallback) shipped `input: toolCall.args` uncoerced — Converse rejects a non-object `toolUse.input`.
2. **Unmirrored inline blocks throw.** An Anthropic-shaped inline `tool_use` block whose id is absent from `tool_calls` keeps the raw streamed JSON string as its input; the formatter threw `Invalid Anthropic tool_use content block` before the request was even built — failing the whole turn on a recoverable shape.

## Fix

New `coerceBedrockToolUseInput` (deliberate duplicate of the Anthropic fork's `coerceAnthropicToolUseInput`, so each fork stays self-contained against its upstream): a string parses when it forms a complete JSON object; every other non-object shape degrades to `{}` — the call already executed, so the replayed input is informational.

Applied at all four emission sites:
- `convertAIMessageToConverseMessage`: unmirrored inline blocks coerce instead of throwing (blocks missing `id`/`name` still throw — nothing to build a `toolUse` from); `tool_calls` materialization coerces args.
- `convertFromV1ToChatBedrockConverseMessage`: v1 `tool_call` block path and `tool_calls` fallback both coerce.

## Tests

New `convertToConverseMessages — non-object toolUse input coercion` suite in `message_inputs.test.ts`:
- null inline input + null args → `toolUse.input: {}`
- unmirrored inline block with a complete JSON-object string parses instead of throwing
- partial-JSON string, `'123'`, and `null` inline inputs degrade to `{}`
- a block missing `id`/`name` still throws
- v1 `tool_call` block and `tool_calls` fallback both coerce null args

All `src/llm/bedrock/utils/` suites green (38 tests).

### DIFF
--- a/src/llm/bedrock/utils/message_inputs.test.ts
+++ b/src/llm/bedrock/utils/message_inputs.test.ts
@@ -79,6 +79,120 @@ describe('convertToConverseMessages — Anthropic tool replay', () => {
   });
 });
 
+/**
+ * Bedrock Converse requires `toolUse.input` to be a JSON object. History can
+ * carry non-object values: pre-3.x context-pressure truncation persisted
+ * `null` onto both a block's inline input and its `tool_calls` args (see the
+ * Anthropic-replay fix in PR #369), and Anthropic-shaped inline blocks keep
+ * the raw streamed JSON string. These must coerce — never ship as-is, never
+ * throw the whole request away.
+ */
+describe('convertToConverseMessages — non-object toolUse input coercion', () => {
+  const toolUseBlocks = (result: ConverseResult): ConverseBlock[] =>
+    assistantContent(result).filter((b) => b.toolUse != null);
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #373, opened 2026-07-31
fix: bound traced media and transient OpenAI retries

AlexZahar:
## Summary

- removes image bytes and signed URLs from Langfuse chat-model input/output attributes before OTEL serialization while preserving bounded media type/byte-count metadata
- clones callback-only messages so provider inputs/outputs are never mutated
- replaces the existing LangChain `AsyncCaller` policy on OpenAI/Azure wrappers with an exact transient classifier
- retries only 408, 429, 5xx, recognized network failures, and fetch failures
- honors `Retry-After` up to 60 seconds; aborts, quota exhaustion, invalid payload/content-filter responses, and other 4xx failures are terminal

This remains one retry layer: the OpenAI SDK clients are already configured with native `maxRetries: 0`; the existing LangChain caller still supplies bounded exponential backoff and jitter.

## Production evidence

During the 2026-07-31 Agentic Karl Datadog EU audit, image requests produced 6 `OTEL request body exceeds 16MB` warnings, 12 `OTEL oversized span detected` warnings, and transient Foundry 408/5xx responses. Invalid-payload 400s must remain non-retryable.

## Verification

- `src/specs/langfuse-callbacks.test.ts` + `src/llm/openai/retry.test.ts`: 43 tests passed
- `npm run build`
- focused ESLint passed (one pre-existing warning remains at `src/llm/openai/index.ts:468`)
- touched-file import-order check passed

## Rollout

Publish the merged SDK version, then bump Agentic Karl's `@librechat/agents` dependency and lockfile. After deployment, verify zero new oversized OTEL warnings and compare Foundry status counts with terminal `result:error` generation metrics. No live provider or observability state was changed.

### DIFF
--- a/src/langfuse.ts
+++ b/src/langfuse.ts
@@ -10,6 +10,8 @@ import {
 import type {
   AIMessageChunkFields,
   AIMessageFields,
+  BaseMessage,
+  MessageContent,
   UsageMetadata,
 } from '@langchain/core/messages';
 import type {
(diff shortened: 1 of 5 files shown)


upman:
@AlexZahar Can we make the file size and this feature configurable? There are cases where we want to keep the file in tracing

Also curious about this warning "OTEL request body exceeds 16MB".  Where is it coming from?

PR #377, opened 2026-08-01
🔌 feat(search): Allow Callers to Inject an HTTP(s) Agent for Outbound Requests

dustinhealy:
## Summary

I add optional `httpAgent` and `httpsAgent` fields to the web-search tool config and apply them to every outbound request the search, scrape, and rerank clients make, so a host application can constrain outbound connection policy (connection pooling, egress rules, resolved-IP validation) on this traffic. It is opt-in and defaults to current behavior.

- Add `httpAgent`/`httpsAgent` to `SearchToolConfig` (`src/tools/search/types.ts`) and thread them through the per-provider factories in `tool.ts` and `search.ts` and the reranker in `rerankers.ts`.
- Apply both agents on every outbound axios call under `src/tools/search/`: firecrawl, serper, tavily (scraper and search), keenable (scraper and search), fastCRW (scraper and search), the serper and searxng search APIs, and the jina and cohere rerankers.
- Pass both agents on every call so a protocol-changing redirect still uses a caller-controlled agent, since `follow-redirects` selects the http or https agent per hop.
- Keep it fully backward compatible: both fields are optional and default to undefined, so with no agent supplied axios behaves exactly as before.

## Change Type

## Testing

- `npm run build` (tsdown + tsc): clean.
- `npx jest src/tools/search`: 156 passed across 10 suites, including 7 new cases in `src/tools/search/http-agent.test.ts` that cover each threading shape (scraper config object, reranker config object, options-bag search factory, internal-config Serper POST and SearXNG GET) plus backward-compat guards asserting no agent is attached when the fields are omitted.
- `npx eslint "src/**/*.ts"`: clean.

### DIFF
--- a/src/tools/search/tool.ts
+++ b/src/tools/search/tool.ts
@@ -452,6 +452,8 @@ export const createSearchTool = (
     jinaApiKey,
     jinaApiUrl,
     cohereApiKey,
+    httpAgent,
+    httpsAgent,
     onSearchResults: _onSearchResults,
     onGetHighlights,
   } = config;
@@ -498,6 +500,8 @@ export const createSearchTool = (
     crwApiKey,
     crwApiUrl,
     crwSearchOptions,
+    httpAgent,
+    httpsAgent,
   });
 
   /** Create scraper based on scraperProvider */
@@ -508,6 +512,8 @@ export const createSearchTool = (
       ...serperScraperOptions,
       apiKey: serperApiKey,
       timeout: scraperTimeout ?? serperScraperOptions?.timeout,
+      httpAgent,
+      httpsAgent,
       logger,
     });
   } else if (scraperProvider === 'tavily') {
@@ -519,6 +525,8 @@ export const createSearchTool = (
         process.env.TAVILY_API_KEY,
       apiUrl: tavilyScraperOptions?.apiUrl ?? tavilyExtractUrl,
       timeout: scraperTimeout ?? tavilyScraperOptions?.timeout,
+      httpAgent,
+      httpsAgent,
       logger,
     });
   } else if (scraperProvider === 'crw') {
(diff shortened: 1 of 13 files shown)


dustinhealy:
@codex review

src/tools/search/tool.ts
```diff
@@ -508,6 +512,8 @@ export const createSearchTool = (
       ...serperScraperOptions,
       apiKey: serperApiKey,
       timeout: scraperTimeout ?? serperScraperOptions?.timeout,
+      httpAgent,
+      httpsAgent,
```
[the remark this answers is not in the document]
dustinhealy: Fixed in `bf0c2267`. The scraper and Tavily/Keenable/CRW search constructors now fall back to the option-bag `httpAgent`/`httpsAgent` when no top-level agent is set; a top-level agent still overrides, matching the existing `timeout`/`apiKey` fallback pattern in the same literals.

dustinhealy:
@codex review

src/tools/search/types.ts
```diff
@@ -287,15 +296,15 @@ export interface FirecrawlConfig {
   firecrawlOptions?: FirecrawlScraperConfig;
 }
 
-export interface SerperScraperConfig {
+export interface SerperScraperConfig extends HttpAgentConfig {
```
danny-avila: Make a generic Search config, not just limited to HttpAgentConfig, and consolidate repeated config fields if it makes sense

danny-avila (changes requested):

src/tools/search/types.ts
dustinhealy: Fixed in `85d2843e`. Added `BaseSearchProviderConfig` in `src/tools/search/types.ts` (`extends HttpAgentConfig` plus `apiKey`/`apiUrl`/`timeout`/`logger`) and repointed the five scraper configs at it, dropping the duplicated quartet from each. Kept `HttpAgentConfig` as the agents-only mixin for the search-option bags (`TavilySearchOptions`/`CrwSearchOptions`/`KeenableSearchOptions`), which share the agents and `timeout` but not `apiKey`/`apiUrl`/`logger`, and for `SearchConfig`, whose credentials are provider-prefixed. The `Omit<...>` scrape-option helper types are unchanged since `Omit` resolves inherited keys. Left Keenable's `apiUrl` doc comment inline, and left the Crw/Firecrawl browser-field overlap alone for now (can pull it into a sub-base if you want it).

danny-avila (approved):

PR #380, opened 2026-08-02
🩹 fix: Guard Optional input_tokens_details in Responses Cache-Write Extraction

thobai:
## Summary

`getCacheWriteTokens` / `attachCacheWriteMetadata` in `src/llm/openai/index.ts` read `usage.input_tokens_details.cache_write_tokens` with only the outer `?.` (on `usage`) — safe against OpenAI itself, which always populates `input_tokens_details`, but **any OpenAI-compatible server that omits it** throws:

```
TypeError: Cannot read properties of undefined (reading 'cache_write_tokens')
```

on every single completion. Hit this running LibreChat against a local `mlx_vlm.server` backend (any OpenAI-compatible server without OpenAI's exact `usage` shape hits the same crash) — every agent response failed with `[api/server/controllers/agents/client.js #sendCompletion] Unhandled error type Cannot read properties of undefined (reading 'cache_write_tokens')`.

## Root cause

`ResponsesUsageWithCacheWrite` is a plain intersection with the SDK's `ResponseUsage` type:

```ts
type ResponsesUsageWithCacheWrite = OpenAIClient.Responses.ResponseUsage & {
  input_tokens_details?: OpenAIClient.Responses.ResponseUsage['input_tokens_details'] & {
    cache_write_tokens?: number;
  };
};
```

The SDK declares `input_tokens_details` **required** (true for OpenAI itself), so the intersection keeps it required in the merged type despite the local `?:` redeclaration — TypeScript intersections don't let a narrower member relax a wider one's requiredness. This is also why a bare `?.` fix trips `@typescript-eslint/no-unnecessary-condition`: the type system is (internally-consistently, if incorrectly for this use case) certain the field can't be missing.

## Fix

- Added the missing `?.` before `.cache_write_tokens` in both `getCacheWriteTokens` and `attachCacheWriteMetadata`.
- Switched `ResponsesUsageWithCacheWrite` to the same `Omit`-then-redeclare pattern this file already uses for `CompletionUsageWithCacheWrite` (the analogous Completions API field, see `deepseek.test.ts`), so the type honestly reflects that the field is optional for non-OpenAI backends — this resolves the lint warning at its actual source instead of just silencing it.
- Exported the three helper functions (previously module-private) for direct unit test coverage.

## Test plan
- [x] New `src/llm/openai/cacheWriteTokens.test.ts`: 7 cases covering missing `input_tokens_details`, missing `usage` entirely, the field-present success path, and the serialized-metadata fallback, across all three functions.
- [x] Verified the new tests fail with the original `TypeError` when the fix is reverted (confirms they actually catch the regression, not just exercise the code path).
- [x] `npx jest src/llm/openai/` — 171/171 pass, no regressions.
- [x] `npx eslint src/llm/openai/index.ts src/llm/openai/cacheWriteTokens.test.ts` — clean, no warnings (the `no-unnecessary-condition` warning from the `?.`-only version of this fix is gone with the `Omit` type change).
- [x] `npx tsc --noEmit` — clean.

### DIFF
--- a/src/llm/openai/cacheWriteTokens.test.ts
+++ b/src/llm/openai/cacheWriteTokens.test.ts
@@ -1,0 +1,112 @@
+import { AIMessage } from '@langchain/core/messages';
+import type { OpenAIClient } from '@langchain/openai';
+
+import {
+  getCacheWriteTokens,
+  attachCacheWriteUsage,
+  attachCacheWriteMetadata,
+} from './index';
+
+/**
+ * Regression coverage for a crash reported against OpenAI-compatible
+ * third-party servers (e.g. mlx_vlm.server) whose `/v1/responses` usage
+ * payload omits `input_tokens_details` entirely — a shape the OpenAI API
+ * itself always populates, but which `ResponsesUsageWithCacheWrite`
+ * declares optional. Reading `.cache_write_tokens` off that missing object
+ * without a second `?.` threw "Cannot read properties of undefined
+ * (reading 'cache_write_tokens')" on every completion from such a server.
+ */
+describe('cache write token extraction (Responses API)', () => {
+  describe('getCacheWriteTokens', () => {
+    it('returns undefined without throwing when usage has no input_tokens_details', () => {
+      const message = new AIMessage({
+        content: 'hi',
+        response_metadata: {
+          usage: {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #382, opened 2026-08-04
📝 docs: Strengthen /mnt/data Persistence Steering in Sandbox Tool Guidance

danny-avila:
## Summary

A production session showed a model writing helper files to `/tmp` in one `bash_tool` call and reading them in the next, despite the guidance already documenting that `/tmp` is same-call scratch. The documentation existed but did not steer: the persistence rule sat in a trailing clause, the extension examples covered only ten data formats (.json/.txt/.csv/.tsv/.log/.parquet/.png/.jpg/.pdf/.xlsx), and everything else was dismissed as "odd extensions."

That narrow list is also wrong in a way that discourages the correct pattern. The code API's output-collection filter accepts all common source, text, data, document, image, and archive formats (~100 extensions including .py/.sh/.sql/.md), so writing a helper module to /mnt/data/lib.py and importing it in later calls works and is exactly what a model should do for multi-call workflows. The old wording implied it would not persist.

## Changes

`CODE_ARTIFACT_PATH_GUIDANCE` (interpolated into the code, bash, and PTC tool descriptions) now:
- Leads with the imperative rule: anything a later call needs must be written under `/mnt/data` in the same call that produces it; `/tmp` never survives the call.
- Gives representative extension examples grounded in the actual output filter (.py/.sh/.sql/.md/.json/.csv/.parquet/.png/.pdf/.zip and similar) and states the real rule: extensionless or unrecognized extensions are not kept.
- Keeps the failed-execution warning, phrased as an action: failed executions register nothing; fix and rerun.

The reactive output reminder (`appendTmpScratchReminder`) already fires on all three executor paths and is unchanged; this PR sharpens the plan-time text those reminders back up.

## Tests

Two literal-substring assertions updated to the new phrasing (`BashExecutor.test.ts`, `ProgrammaticToolCalling.test.ts`). Full `src/tools/__tests__` suite: 936 passed.

### DIFF
--- a/src/tools/CodeExecutor.ts
+++ b/src/tools/CodeExecutor.ts
@@ -23,7 +23,7 @@ export const emptyOutputMessage =
   'stdout: Empty. Ensure you\'re writing output explicitly.\n';
 
 export const CODE_ARTIFACT_PATH_GUIDANCE =
-  'Persist handoff artifacts in `/mnt/data` with standard extensions (.json/.txt/.csv/.tsv/.log/.parquet/.png/.jpg/.pdf/.xlsx); failed executions do not register new files; `/tmp` and odd extensions are same-call scratch only, not later-call storage.';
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #386, opened 2026-08-05
fix: Prevent unsafe token sentinels from overflowing context

danny-avila:
## Summary

- treat the bounded serializer's unknown-size sentinel as unknown rather than real token usage
- count proxy-backed structured tool-call arguments through their bounded provider-safe representation without invoking getters
- reject invalid, non-safe, sentinel, aggregate-overflow, and post-Claude-correction token measurements
- report genuinely unmeasurable message, content, and metadata branches as machine-readable `unsafe_token_measurement` errors
- add regression coverage for proxy traps, accessors, traversal limits, Claude correction, diagnostics, and safe proxy tool-call history

## Root cause

`serializeStructuredValueBounded` uses `Number.MAX_SAFE_INTEGER` to signal that exact serialized length is unknown after a defensive inspection or traversal limit. Token accounting extrapolated from that marker as though it were a real character count. Claude's 1.1 correction then surfaced the sentinel as roughly 9.9 quadrillion tokens and caused a false context overflow before any provider call.

## Impact

Unsafe inspection sentinels no longer become fabricated context usage. Supported structured proxy histories remain bounded and countable, while branches that cannot be safely inspected fail distinctly from genuine context-window overflows and identify only the offending path. A final invariant prevents future estimators or provider corrections from leaking unsafe counts through this API.

## Validation

- `npx jest src/specs/tokens.test.ts src/graphs/__tests__/Graph.contextOverflow.test.ts --runInBand` — 94 tests passed
- `npx tsc --noEmit`
- `npm run build`
- `npm run sort-imports:check` — changed files pass; repository check reports 17 pre-existing unrelated files

### DIFF
--- a/src/specs/tokens.test.ts
+++ b/src/specs/tokens.test.ts
@@ -72,6 +72,32 @@ describe('createTokenCounter with different encodings', () => {
     expect(count).toBeGreaterThan(0);
   });
 
+  test('claude correction keeps proxy-backed tool history safely measurable', async () => {
+    let getCalls = 0;
+    const args = new Proxy(
+      { query: 'select * from reports' },
+      {
+        get(target, property, receiver) {
+          getCalls++;
+          return Reflect.get(target, property, receiver);
+        },
+      }
+    );
+    const message = {
+      content: '',
+      tool_calls: [{ id: 'proxy-call', name: 'query', args }],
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #406, opened 2026-08-11
feat: add explicit host provider registration seam

vb47w92g7g-wq:
## Summary

- add an explicit registry for host-provided chat model constructors
- preserve the existing typed built-in provider lookup
- resolve built-in and host providers through `initializeModel`
- export the registration and lookup APIs
- add registry behavior tests

## Design notes

This is a generic provider-extension seam only. It does not add an ERIS dependency, register any provider implicitly, modify AgentClient, or enable any runtime behavior by default.

Host applications must explicitly call `registerChatModelProvider`. Existing built-in provider names cannot be overwritten.

## Validation

The added test covers registration, duplicate rejection, invalid inputs, built-in lookup preservation, and unsupported-provider errors.

ERIS integration remains downstream and feature-gated.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,12 +52,6 @@ export type * from './types';
 /* LangChain compatibility facade */
 export * from './langchain';
 
-/**
- * HITL primitives re-exported from `@langchain/langgraph` so hosts that
- * build durable checkpoint savers, dispatch `Command({ resume })`, or
- * detect interrupts can do so against the same langgraph instance the
- * SDK was compiled against — avoiding accidental dual-version drift.
- */
 export {
   Command,
   INTERRUPT,
@@ -76,7 +70,12 @@ export type {
   OpenRouterReasoningEffort,
   ChatOpenRouterCallOptions,
 } from './llm/openrouter';
-export { getChatModelClass } from './llm/providers';
+export {
+  getChatModelClass,
+  getRegisteredChatModelClass,
+  registerChatModelProvider,
+} from './llm/providers';
+export type { RegisteredChatModelConstructor } from './llm/providers';
 export { CustomChatMistralAI } from './llm/mistral';
 export {
   smoothStream,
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
Thank you, @vb47w92g7g-wq — the core intention here is sound: provider construction should be host-extensible instead of requiring an agents release for every compatible adapter. I used this PR as the foundation for #458 and cherry-picked all four commits so the original authorship and contribution history are preserved.

I did not merge this head directly because the intervening SDK changes exposed a few contract gaps, including the four open Codex findings, which I confirmed were valid:

- A registered model was assumed to support `bindTools`, so a tool-enabled run could fail with an opaque method error.
- The public configuration types still narrowed `initializeModel` to built-in option shapes, preventing real host-specific constructor options from flowing through the normal seam.
- Object-backed lookup treated inherited names such as `constructor` and `__proto__` as registrations.
- Runtime callers could register a callable but non-constructible function and defer failure until model initialization.

The replacement keeps your explicit runtime registration idea and extends it into a single provider-registration module. Built-ins and hosts now use the same `Map`-backed registry; registrations also own provider family, manual tool-stream behavior, and strict message alternation so behavior does not remain scattered in separate static sets. Duplicate names fail closed, constructors are checked up front, tool binding gets an explicit provider-specific guard, and the returned disposer is ownership-scoped for safe tests/hot reload.

For host typing, #458 adds a stable `@librechat/agents/provider-registration` package subpath and a declaration-mergeable `CustomProviderOptionsMap`. That carries host options through model initialization while preserving precise built-in mappings. The process-local lifecycle and the boundary between family behavior and exact built-in behavior are documented in ADR 0003 and the README.

Validation on the replacement currently includes a clean build and type-check, no runtime dependency cycles, and 100 focused tests across registration, provider request shaping, invocation, alternation, and fallbacks. I’m closing this PR as superseded by #458, with your contribution retained in that branch. Thanks again for identifying and implementing the right initial seam.

vb47w92g7g-wq:
Danny,

I hope this comment finds you well. Thank you for taking the time to review this PR and to build on it. I'm honored that you saw value in my initial host-provider registration seam and have carried the idea forward into a more robust implementation. This feature in `agents` will be quite helpful for my genealogy project; its inclusion means a great deal to me.

PR #412, opened 2026-08-14
🚫 fix: Drop Model-Turn Prefill for Gemini 3.7 Flash

danny-avila:
## Summary

Google's prefill restriction — a request whose `contents` end with a `model`-role turn returns HTTP 400 — applies to **Gemini 3.7 Flash** exactly as it does to Gemini 3.6 Flash and Gemini 3.5 Flash-Lite. `NO_PREFILL_GEMINI_MODELS` is model-enumerated, so 3.7 Flash was not covered and prefill flows would fail against it.

Adds `gemini-3.7-flash` to the list.

## Why it matters

A trailing `model` turn is produced by prefill flows — most commonly editing an assistant reply and resubmitting. Without this entry, those requests reach Gemini 3.7 Flash unmodified and return a 400 instead of being repaired the way they already are for 3.6 Flash.

Companion to danny-avila/LibreChat#14802, which adds first-class Gemini 3.7 Flash support on the LibreChat side. That change is independent of this one — LibreChat's model list, pricing, and Flash-family parameter handling do not depend on this fix, but prefill on 3.7 Flash stays broken until this ships.

## Changes

- `src/llm/google/utils/common.ts` — add `gemini-3.7-flash` to `NO_PREFILL_GEMINI_MODELS`; update the doc comment.
- `src/llm/google/utils/common.test.ts` — cover the model and its versioned aliases (`models/gemini-3.7-flash-latest`) in `rejectsModelTurnPrefill`, and add a `dropUnsupportedModelTurnPrefill` case.

## Testing

`npx jest src/llm/google/utils/common.test.ts` — 11 passed.

Verified the new assertions are real: reverting `common.ts` alone turns both new tests red (`rejectsModelTurnPrefill › is true for models that reject a trailing model turn` and `dropUnsupportedModelTurnPrefill › drops a trailing model turn for Gemini 3.7 Flash`).

`npx tsc --noEmit` — clean.

Note: `src/llm/google/llm.spec.ts` fails to run in my sandbox for want of a `GOOGLE_API_KEY` in the process env. Confirmed pre-existing — it fails identically at `origin/main` with this change reverted.

## Reference

- https://ai.google.dev/gemini-api/docs/latest-model#api-changes-and-parameter-updates

### DIFF
--- a/src/llm/google/utils/common.test.ts
+++ b/src/llm/google/utils/common.test.ts
@@ -116,9 +116,11 @@ describe('convertResponseContentToChatGenerationChunk seal metadata', () => {
 
 describe('rejectsModelTurnPrefill', () => {
   test('is true for models that reject a trailing model turn', () => {
+    expect(rejectsModelTurnPrefill('gemini-3.7-flash')).toBe(true);
     expect(rejectsModelTurnPrefill('gemini-3.6-flash')).toBe(true);
     expect(rejectsModelTurnPrefill('gemini-3.5-flash-lite')).toBe(true);
     expect(rejectsModelTurnPrefill('models/gemini-3.6-flash')).toBe(true);
+    expect(rejectsModelTurnPrefill('models/gemini-3.7-flash-latest')).toBe(true);
     expect(rejectsModelTurnPrefill('google/gemini-3.5-flash-lite-latest')).toBe(
       true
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #421, opened 2026-08-16
🐛 fix: Close A Self-Limited Child's Steps As `failed`, Not `cancelled`

danny-avila:
## Summary

Follow-up to #417, from a structured review of the v3.6.0 run-step lifecycle work. Two commits: a status-mislabel fix and a stale-contract doc removal.

### The fix (`aa26ef46`)

A child subagent that trips **its own** stream limit had its run steps closed as `cancelled`. The catch block self-aborts the child breaker *before* the closure sweep reads the composed signal:

```ts
if (error instanceof StreamLimitExceededError) {
  childBreaker.abort(error);        // flips childSignal.aborted
}
await this.closeChildRunSteps(
  childGraph,
  childSignal.aborted ? 'cancelled' : 'failed',   // reads it after → always 'cancelled'
  childTerminalAt
);
```

So the child's own failure reported as an intentional stop — while the parent, whose `resolveSweepStatus` this code is documented to mirror, stamps `failed` for the same incident. One error, contradictory terminal statuses; and in the LibreChat UI that consumes these statuses (#14871/#14873), a genuine limit failure renders as a user cancellation.

The status now derives from a snapshot of `childSignal.aborted` captured at catch entry, before the self-abort. The intentional cases are preserved: a trip arriving from a parallel sibling (or a caller abort) has already aborted the composed signal by the time the error lands, so those still read `cancelled`.

### The docs commit (`c5b7287a`)

`closeRunStep`'s JSDoc promised a `restamp` close that refreshes a completed TOOL_CALLS step's `completed_at` in the eager-execution race — but `RunStepCloseOptions` carries no such option and the implementation unconditionally refuses any second close; the mechanism was dropped as unreachable in the PR that introduced the doc. Removed so nobody designs around behavior that never happens.

## Testing

Two new tests in `SubagentExecutor.test.ts` pin both bounds:

- **`closes child steps as 'failed' when the child trips its own stream limit`** — fails against the previous code with exactly the mislabel (`Expected: "failed" / Received: "cancelled"`), verified by stashing the fix and re-running.
- **`closes child steps as 'cancelled' when the abort preceded the error`** — the batch breaker aborts mid-flight before the child rejects; the intentional-stop path keeps reading `cancelled` under the snapshot.

Full file suite: **109/109**. `npx tsc --noEmit` clean; eslint/prettier/import-sort clean via the commit hooks.

Related issues filed from the same review, not addressed here: AI-1811 (closed-handler dispatch isolation), AI-1812 (`AgentSession` closeStep duplication), AI-1813 (non-streaming final-step race).

https://claude.ai/code/session_014vLhxCFMYkCaTsoFTiAjJ5

---
_Generated by [Claude Code](https://claude.ai/code/session_014vLhxCFMYkCaTsoFTiAjJ5)_

### DIFF
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -161,6 +161,7 @@ import { createCloudflareCodingToolBundle } from '@/tools/cloudflare';
 import { buildSubagentToolParams } from '@/tools/SubagentTool';
 import { initializeLangfuseTracing } from '@/instrumentation';
 import { shouldTriggerSummarization } from '@/summarization';
+import { isRunStepResumeState } from '@/tools/runStepResume';
 import { resolveLocalToolsForBinding } from '@/tools/local';
 import { createSummarizeNode } from '@/summarization/node';
 import { messagesStateReducer } from '@/messages/reducer';
@@ -168,7 +169,6 @@ import { createSchemaOnlyTools } from '@/tools/schema';
 import { AgentContext } from '@/agents/AgentContext';
 import { createFakeStreamingLLM } from '@/llm/fake';
 import { handleToolCalls } from '@/tools/handlers';
-import { isRunStepResumeState } from '@/tools/runStepResume';
 import { isThinkingEnabled } from '@/llm/request';
 import { resolveMaxSeals } from '@/llm/preempt';
 import { initializeModel } from '@/llm/init';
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

---

PR #422, opened 2026-08-17
🔐 fix: Seal Responses Reasoning Items Per Item On Interrupt

danny-avila:
## Summary

Interrupting a `gpt-5.6` turn that emitted **more than one reasoning item** makes the **next** request fail with:

```
400 The encrypted content for item rs_... could not be verified.
Reason: Encrypted content could not be decrypted or parsed.
```

**Verified live against `gpt-5.6-terra`** — reproduced and then fixed end to end (details below).

## When this fires

A single Responses turn emits multiple reasoning items whenever a **server-side tool** interleaves with reasoning — `web_search`, `code_interpreter`, `image_generation`. Live, one `web_search` request produced ~10 reasoning items in one turn:

```
added:reasoning:rs_…be64  → done:reasoning:rs_…be64
added:web_search_call:ws_…c063 → done:web_search_call:ws_…c063
added:reasoning:rs_…c7de  → done:reasoning:rs_…c7de
… (repeats) …
added:message:msg_…fffe  → *** interrupt lands here ***
```

A plain no-tool turn emits a single reasoning item (confirmed live) and is unaffected, no matter how long the answer is. Multiple *summary parts* within one item are also fine — it is multiple **items** that break.

## Root cause

An interrupted turn never receives `response.completed`, so it has no authoritative `response_metadata.output` and replays from the single `additional_kwargs.reasoning` slot instead.

That one slot has to stand in for a turn that can emit **many** reasoning items, and the chunk merge folds it field by field:

| field | merge result across N items |
|---|---|
| `id` | last wins |
| `status` | concatenated — `"completedcompleted…"` |
| `encrypted_content` | concatenated — every item's ciphertext welded together |

The replay then hands the provider one item id paired with a blob that is not that item's ciphertext. Live, the welded blob measured **27,012 chars** against ~1,700 for a single item.

Completed turns are unaffected: the authoritative terminal output wins over the slot. Claude models are unaffected: Anthropic stores `encrypted_content` per content block and assigns rather than appends.

The nastiest shape is an interrupt landing *inside* the second item — `status` stays a single legal `'completed'` while `encrypted_content` still belongs to the first item. That passes every local guard including `hasReplayableEncryptedReasoning`; only the provider catches it.

## Fix

- Track the reasoning item currently streaming so a terminal `encrypted_content` seals against **its own** id.
- Keep the slot describing whichever item most recently sealed, so `id`, ciphertext, and `status` always come from the same item.
- Items that never completed no longer replay a bare, unresolvable id.

## Live verification

Real `gpt-5.6-terra`, `store: false`, `include: ['reasoning.encrypted_content']`, `web_search` bound, interrupted mid-answer, then the accumulated message replayed to a real follow-up request:

| | accumulated `encrypted_content` | replay result |
|---|---|---|
| before | 27,012 chars (many items welded) | `400 … could not be verified` — **reported error reproduced** |
| after | 3,428 chars (one intact item) | **accepted**, model answered normally |

## Tests

`src/llm/openai/reasoning-replay.spec.ts` (new) drives the real stream pipeline across three interrupt shapes and asserts exactly one intact item replays, paired with its own id.

`src/specs/preemptSeal.test.ts` gains an end-to-end interrupt-then-steer case that inspects the actual resumed provider payload.

Both were verified to fail without the fix — the steer test sends `rs_inflight` carrying `rs_sealed`'s ciphertext.

Full suite: 4225 passing. The only failures (`anthropic`/`google`/`vertexai` `llm.spec.ts`) fail identically on `main` — credential-gated live tests.

## Not fixed upstream

Checked against latest published, not our pins (we are on `@langchain/openai@1.5.5` / `@langchain/core@1.2.3`):

| package | latest | relevant behavior |
|---|---|---|
| `@langchain/openai` | 1.5.8 | still returns `null` for `response.output_item.done` on a reasoning item, so our own `convertDroppedResponsesReplayOutput` fallback is what writes `encrypted_content` into the slot |
| `@langchain/core` | 1.2.8 | chunk merge still string-concatenates |

A dependency bump does not address this, and upstream never sees the field, so the per-item boundary belongs here.

### DIFF
--- a/src/llm/openai/index.ts
+++ b/src/llm/openai/index.ts
@@ -43,6 +43,7 @@ import type { SeenScalarMetadata } from './streamMetadata';
 import type { HeaderValue, HeadersLike } from './types';
 import type { PromptCacheTtl } from '@/messages/cache';
 import {
+  OPENAI_RESPONSES_ACTIVE_REASONING_ID_KEY,
   OPENAI_RESPONSES_REPLAY_POSITIONS_KEY,
   projectOpenAIResponsesToolMessageContent,
   projectToolStreamContentForProvider,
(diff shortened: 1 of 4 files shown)


danny-avila:
## Not fixed upstream — checked

Verified against the **latest published** versions, not our pins (we're on `@langchain/openai@1.5.5` / `@langchain/core@1.2.3`):

| package | latest | relevant behavior |
|---|---|---|
| `@langchain/openai` | 1.5.8 | `convertResponsesDeltaToChatGenerationChunk` still returns `null` for `response.output_item.done` on a reasoning item, so our `convertDroppedResponsesReplayOutput` fallback is still the thing writing `encrypted_content` into the slot |
| `@langchain/core` | 1.2.8 | the chunk merge still string-concatenates — same three-item input still yields `status: "completedcompleted"` and `encrypted_content: "ENC_AENC_B"` |

So a dependency bump does not address this, and there's nothing to wait on upstream.

Worth noting the single-slot-for-N-items design is ours either way: `additional_kwargs.reasoning` only holds `encrypted_content` because our own fallback puts it there, so the per-item boundary belongs in this repo regardless of what upstream does later.

danny-avila:
@codex review

danny-avila:
## Live verification against the real provider

Everything above was re-verified end to end on real `gpt-5.6-terra` (`store: false`, `include: ['reasoning.encrypted_content']`, `web_search` bound), interrupting mid-answer and replaying the accumulated message to a real follow-up request.

| | accumulated `encrypted_content` | replay result |
|---|---|---|
| fix disabled | 27,012 chars | `400 The encrypted content for item rs_00b1d957… could not be verified` |
| fix enabled | 3,428 chars | accepted, model answered normally |

A single reasoning item's ciphertext measures ~1,700 chars, so the 27k blob was roughly sixteen items welded together.

Two corrections this forced to the original description, worth calling out for reviewers:

1. **The trigger is server-side tools, not long output.** A plain no-tool turn emits exactly one reasoning item however long the answer runs, and never hits this. It takes `web_search` / `code_interpreter` / `image_generation` interleaving with reasoning to put multiple items in one turn.
2. **Multiple summary *parts* are not the bug.** One item with three summary parts is fine. Langfuse renders each part as its own reasoning block, which is what originally made a single-item turn look like a three-item one.

PR #426, opened 2026-08-17
🪢 feat: Support Custom Headers On Langfuse Exports

danny-avila:
## Summary

Self-hosted Langfuse instances sitting behind an authenticating proxy or API gateway (Cloudflare Access, oauth2-proxy, a corporate gateway wanting `X-Api-Key`) need extra HTTP headers on every export request. There was no way to supply them.

`@langfuse/otel` already accepts `additionalHeaders` and applies it to **both** export surfaces — the OTLP trace exporter and the `LangfuseAPIClient` behind media uploads — so this is purely a passthrough that was never wired up.

## Changes

- `LangfuseConfig.additionalHeaders?: Record<string, string>` (`src/types/graph.ts`)
- Passed through in every credential branch of `getLangfuseSpanProcessorParams` — explicit config credentials, env credentials, and the config-baseUrl-with-env-credentials branch. Omitted entirely when unset or empty, so params are unchanged for everyone not using this.
- Merged in `resolveLangfuseConfig` like the other map-valued fields, agent overriding run on key collisions.

## Why headers are part of destination identity

`getLangfuseDestinationKey` gains a header digest. This is the load-bearing decision in the PR:

- A gateway may **route** on these headers, so two configs identical in credentials and baseUrl can be different projects. Excluding headers from the key would let a span inherit an ambient parent bound to another destination — the exact cross-destination parenting the registry exists to prevent.
- `RoutingLangfuseSpanProcessor` caches processors by a key derived from the destination key. Excluding headers would mean a rotated proxy credential silently keeps exporting through the stale cached exporter.

The digest is order- and case-insensitive (header names are case-insensitive in HTTP), so header maps that differ only in key order or casing share one exporter instead of duplicating it. Absent and empty both yield `undefined`, keeping keys byte-identical to today for the no-headers case. Values are hashed, like `secretKey` already is — proxy tokens are credentials.

## Scope note

Deployment-level only, and documented as such on the field. One exporter is shared by every span routed to a destination, so these headers cannot carry per-request or per-user identity the way custom-endpoint headers can.

## Testing

- 10 new assertions across `langfuse-span-registry.test.ts` (passthrough in all three credential branches, omission when unset/empty, separation by header name and by rotated value, stability across key order / name casing / empty maps, and that raw header values never appear in the destination key) and `langfuse-tool-output-tracing.test.ts` (run/agent merge precedence, one-sided merge).
- All 5 Langfuse suites green: 96/96.
- `npx tsc --noEmit` and `npx eslint` clean.
- Pre-existing unrelated failures in `llm/{google,vertexai,anthropic}` and `Graph.reasoning` reproduce identically on a clean `main` checkout (they need provider credentials).

## Follow-up (not in this PR)

While tracing the params path I found that **`mediaUploadEnabled` is a phantom option**: it appears in neither the type definitions nor the runtime of `@langfuse/otel@5.4.1` (`grep -c mediaUploadEnabled dist/index.cjs` → 0). It survives type-checking only because conditional spreads bypass excess-property checks. Anything relying on it to suppress media upload — including LibreChat's central-media-disable path — is currently a no-op. Filing separately.

### DIFF
--- a/src/langfuseConfig.ts
+++ b/src/langfuseConfig.ts
@@ -175,6 +175,44 @@ export function resolveToolOutputTracingConfig(
   };
 }
 
+/**
+ * Merges header maps case-insensitively, keeping the override's casing.
+ *
+ * A plain spread would keep both `X-Proxy-Token` and `x-proxy-token`, and
+ * filling a fetch `Headers` from that record *appends* rather than replaces —
+ * the exporter would send one comma-joined `run-token, agent-token` value, so
+ * the agent override never cleanly wins and a gateway sees a malformed
+ * credential. Matches the case-insensitive identity already used for the
+ * destination key.
+ */
+function mergeAdditionalHeaders(
+  base?: Record<string, string>,
+  override?: Record<string, string>
+): Record<string, string> | undefined {
+  if (base == null && override == null) {
+    return undefined;
+  }
+
+  const merged: Record<string, string> = { ...base };
+  if (override == null) {
+    return merged;
+  }
+
+  const baseKeyByLower = new Map<string, string>(
+    Object.keys(merged).map((key) => [key.toLowerCase(), key])
+  );
+  for (const [key, value] of Object.entries(override)) {
+    const lower = key.toLowerCase();
+    const existingKey = baseKeyByLower.get(lower);
+    if (existingKey != null && existingKey !== key) {
+      delete merged[existingKey];
+    }
+    merged[key] = value;
+    baseKeyByLower.set(lower, key);
+  }
+  return merged;
+}
+
 export function resolveLangfuseConfig(
   runLangfuse?: t.LangfuseConfig,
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
Fixed in `b8f2a59e` — the P1 is real and was an inconsistency inside this PR: `getLangfuseDestinationKey` already lowercases header names when building its fingerprint, so two configs differing only in casing resolve to one destination, while the merge kept both entries. Filling a fetch `Headers` from that record appends rather than replaces, so the exporter would have sent a single comma-joined `run-token, agent-token` value and the documented agent override would never cleanly win.

`mergeAdditionalHeaders` now merges case-insensitively and keeps the override's casing, mirroring LibreChat's `mergeHeaders`. New test asserts an agent `x-proxy-token` fully replaces a run `X-Proxy-Token` rather than coexisting with it.

@codex review

PR #431, opened 2026-08-18
🛟 fix: Propagate Responses Stream Errors

### LINKED ISSUE #429: stream errors are silently dropped and result in an empty assistant message
I’m not familiar with front-end development, but I was able to get Codex to fix the issue, and I’ve confirmed that it’s resolved. Here’s what Codex wrote:

---

### Description

  When using `@librechat/agents` with the OpenAI Responses API in streaming mode, an upstream error can be silently ignored.

  For example, the upstream provider returns an error payload like:

  ```json
  {
    "error": {
      "type": "service_unavailable_error",
      "code": "server_is_overloaded",
      "message": "Our servers are currently overloaded. Please try again later.",
      "param": null
    }
  }
```

  Instead of propagating the error to the caller, the agent run is treated as successfully completed. LibreChat emits a final event
  and creates an empty assistant message, so the frontend does not display the upstream error.

  ### Versions

  - LibreChat: v0.8.8-rc1
  - @librechat/agents: 3.6.3
  - API mode: OpenAI Responses API with streaming enabled

  ### Steps to reproduce

  1. Configure an agent to use an OpenAI-compatible endpoint through the Responses API.
  2. Start a streaming generation.
  3. Have the upstream endpoint return a service error during the stream.
  4. Observe that the agent run completes without throwing an error.
  5. Observe that the resulting assistant message has no content.

  ### Actual behavior

  The upstream error event is ignored, the LLM call is reported as complete, and an empty assistant response is finalized.

  ### Expected behavior

  The stream error should be propagated to the caller as an error. The original structured error payload, including type, code,
  message, and param, should remain available so the application can display or handle it appropriately.

### Root cause

  In `convertLibreChatResponsesStream`, every streamed event is passed through `convertResponsesDeltaToChatGenerationChunk` and
  `convertDroppedResponsesReplayOutput`.

  Error events such as `response.failed` are not converted by either function, so `convertedChunk` is `null`. The following branch
  then silently skips the event:

  ```ts
  const convertedChunk =
    convertResponsesDeltaToChatGenerationChunk(event) ??
    convertDroppedResponsesReplayOutput(event);

  if (convertedChunk == null) {
    continue;
  }
```

  Because the error event is skipped instead of thrown, the surrounding catch block is never reached. The stream finishes normally
  with no generated content, causing the run to be finalized as successful with an empty assistant message.

danny-avila:
I fixed Responses API streaming failures so terminal provider errors propagate to callers instead of completing with an empty assistant message.

- Detect `response.failed` and `error` events before unsupported stream events are skipped.
- Preserve structured provider details through the existing OpenAI error path.
- Return an explicit failure when a failed response omits error details.
- Add regression coverage for both terminal error event formats.

Closes #429

## Change Type

## Testing

- `npx jest src/llm/openai/llm.spec.ts --runInBand` — 59 tests passed.
- `npx jest langfuse deterministic-trace-id --runInBand` — 176 tests passed.
- `npx tsc --noEmit` — passed.
- `npx eslint src/llm/openai/index.ts` — passed.
- `git diff --check` — passed.

### **Test Configuration**:

- Node.js v24.16.0
- npm 11.16.0

## Checklist

- [x] I have written tests demonstrating that my changes are effective

### DIFF
--- a/src/llm/openai/index.ts
+++ b/src/llm/openai/index.ts
@@ -58,14 +58,14 @@ import {
   STREAMED_TOOL_CALL_ADAPTER_METADATA_KEY,
   OPENAI_CHAT_SEQUENTIAL_STREAMED_TOOL_CALL_ADAPTER,
 } from '@/tools/streamedToolCallSeals';
-import { isReasoningModel, _convertMessagesToOpenAIParams } from './utils';
-import { INTENT_ARG, isIntentLabelProperty } from '@/tools/intentArg';
-import { smoothStream, resolveStreamDelay } from '@/llm/stream/smoother';
 import {
   hasReasoningKwargs,
   hasToolCallChunks,
   getReasoningKwargsText,
 } from '@/llm/stream/chunkAdapters';
+import { isReasoningModel, _convertMessagesToOpenAIParams } from './utils';
+import { smoothStream, resolveStreamDelay } from '@/llm/stream/smoother';
+import { INTENT_ARG, isIntentLabelProperty } from '@/tools/intentArg';
 import { dropRepeatedScalarMetadata } from './streamMetadata';
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #452, opened 2026-08-23
🧮 perf: Reuse Exact Context Counts Across Provider Requests

danny-avila:
## Summary

I added conservative exact-token reuse across provider requests in an agent run, reducing repeated context-pressure tokenizer work while preserving request-scoped accounting and overflow behavior.

- Retain exact stable-message counts in `AgentContext` through a weak cache while leaving calibration, provider attribution, budgets, and fallback state request-scoped.
- Revalidate every built-in token-relevant field before reuse and force structured content, proxies, accessors, tool calls, and mutations through the exact recount path.
- Keep unmarked custom token counters on the existing uncached path and expose an explicit compatibility marker for deterministic host wrappers such as LibreChat's counter.
- Verify cached and forced-recount measurements match across append, replacement, reordering, mutation, tool metadata changes, and AgentContext resets.
- Add a reproducible real-tokenizer benchmark plus an ADR documenting the correctness and host-lifecycle boundaries.

## Change Type

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Testing

- Ran `npx jest src/llm/contextPressureMeter.test.ts src/agents/__tests__/AgentContext.test.ts --runInBand` (126 tests passed).
- Ran context-overflow, provider-request, tokenizer, and Langfuse focused suites (319 tests passed).
- Ran `npm run build`, `npx tsc --noEmit`, scoped ESLint, and both circular-dependency checks successfully.
- Ran the repository suite: 4,553 tests passed; only three live-provider suites failed because local Anthropic/Google credentials were unavailable and retired Vertex preview models returned 404.
- Ran `npm run bench:context-pressure`; cached and uncached projected totals matched exactly. Ten iterations reduced tokenizer calls from 1,000 to 100 for 100 retained messages and from 5,000 to 500 for 500 retained messages. Median time improved from 8.12 ms to 1.40 ms (5.8x) and from 39.26 ms to 5.92 ms (6.63x).

### **Test Configuration**:

- Node.js 24.16.0
- npm 10.5.2
- macOS arm64

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] I have made pertinent documentation changes
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

PR #456, opened 2026-08-23
⚡ perf: Skip Exact Token Counts the Host Already Supplies

danny-avila:
## Summary

Profiling LibreChat follow-up turns on a 300-message conversation showed **29% of per-turn CPU** in one chain:

`contextPressureMeter.count → getTokenCountForMessage → processValue → isAtomicToolContentBlock → hasUnsafeCallableToJSON → readNativeDate / readNativeArrayBufferByteLength`

Two compounding causes, both fixed here.

### Exact counts were computed and then discarded

`createContextPressureMeter`'s baseline tokenized every retained message up front as `rawTokens`, then — whenever the host had supplied a stored count in `indexTokenCountMap` — used the stored value and dropped the expensive one. `rawTokens` is only ever read as the subtrahend of a projection delta (`rawTokens − baseline[origin].rawTokens`), and that delta is zero by construction when the projected message is the same object as its baseline. The baseline now tokenizes only when no stored count exists, exposes `rawTokens` as a lazy getter for the changed-projection case, and `measure()` skips counting entirely for same-object messages. `forceRawRecount` still takes the unchanged full-recount branch.

The #452 exact-count cache is keyed by message object identity on the per-run `AgentContext`, so it reuses counts across provider requests within a run but cannot hit across turns — every turn builds fresh message objects. That is by design; this change just stops paying for counts nobody reads.

### Native brand checks threw an exception per object

`readNativeDate` called `Date.prototype.getTime.call(value)` on every object and caught the `TypeError`; `readNativeArrayBufferByteLength` did the same with the `byteLength` getter. For the plain objects that make up nearly all message content that is two thrown exceptions — with captured stacks — per object. `node:util/types` `isDate` / `isArrayBuffer` read the same internal slots natively: no user code, no throw, cross-realm correct, and already imported here for `isProxy`. Verified equivalent to the throwing probes on a real Date, a prototype-swapped real Date, a fake with `Date.prototype`, and a plain object; `isArrayBuffer` (not `isAnyArrayBuffer`) preserves the getter's rejection of `SharedArrayBuffer`.

## Measured

LibreChat `main`, Node 24, local MongoDB, in-process fake model, 40 follow-up turns per variant, three alternating rounds:

| conversation | metric | 3.6.14 | this branch |
|---|---|---|---|
| 300 messages | CPU / turn (median) | 58.8 ms | 39.1 ms (−33%) |
| 300 messages | CPU / turn (best stock round vs worst patched) | 44.3 ms | 40.4 ms (−9%) |
| 1000 messages | turn p50 | 76 ms | 56 ms (−27%) |
| 1000 messages | throughput | 13.0/s | 17.7/s (+36%) |

The stock variance across rounds (44 → 63 → 59 ms) is itself the exception/GC pressure; the patched build sat at 37–40 ms. In the windowed CPU profile both probe functions drop out of the top 40 entirely and `@librechat/agents` self-time falls 282 → 108 ms over 20 turns.

## Tests

Three `toHaveBeenCalledTimes` assertions pinned the old tokenizer invocation count and are updated (3 → 2, 101 → 0, 113 → 13 — each the minimum for its scenario). One new test keeps the cross-meter exact-count cache covered for the no-stored-count path (101 calls). `contextPressureMeter`, `toolContent`, and `tokens` suites: 69/69. The provider `llm.spec.ts` suites fail identically on clean `main` (they need live credentials).

`CONTEXT.md` notes the new baseline contract.

Note: the commit is unsigned — GPG wasn't available in the environment that produced it.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -72,6 +72,12 @@ host explicitly declares that they implement the same deterministic surface
 contract.
 See [ADR 0001](docs/adr/0001-reuse-exact-context-token-counts.md) for the
 cache boundary and host-lifecycle trade-offs.
+Host-supplied index token counts are the baseline's accounting weight; a
+retained message is tokenized only when no such count exists or when a
+provider projection replaced its object, since the exact count then serves
+only as the subtrahend of that projection's delta. Native brand checks in
+content validation use `node:util/types` slot reads rather than throwing
+probes, so plain content objects never pay an exception per value.
 
 The meter's estimates decide overflow and recovery only. Provider-reported
 usage remains the source of truth for billing and Langfuse observations.
--- a/src/llm/contextPressureMeter.test.ts
+++ b/src/llm/contextPressureMeter.test.ts
@@ -1,8 +1,4 @@
-import {
-  AIMessage,
-  HumanMessage,
-  ToolMessage,
-} from '@langchain/core/messages';
+import { AIMessage, HumanMessage, ToolMessage } from '@langchain/core/messages';
 import type { BaseMessage } from '@langchain/core/messages';
 import {
   createContextPressureMeter,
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

PR #473, opened 2026-08-27
perf: align compaction with provider cache prefixes

danny-avila:
## Summary

- centralize the live provider tool projection used by normal requests and self-compaction, including local/Cloudflare tool resolution and deferred-tool partitioning
- apply provider-correct Anthropic, OpenRouter, and Bedrock cache markers and TTLs to the compaction tool/history prefixes
- preserve the configured Bedrock model family behind opaque application inference profiles
- treat provider cache-creation/cache-read usage as the receipt instead of predicting eligibility from provider options or environment variables

## Performance proof

The normal model hot path still performs the same single tool-resolution/cache-preparation pass; the work is relocated behind a shared function. Compaction pays one live projection only when it fires.

The Anthropic E2E job now primes an isolated normal-request tool prefix, invokes the former unmarked compaction shape as the baseline, then invokes the aligned shape. It logs before/after cache-read tokens and request latency, asserts that the prime created cache tokens, and requires the aligned request to read more cached input than baseline.

## Verification

- npx tsc --noEmit
- npm run build
- 141 focused cache, graph-overflow, and summarization tests passed; 11 credential-gated tests skipped locally
- changed source lint passes; the summarization live-spec retains six pre-existing no-useless-assignment warnings

## Architecture

ADR 0006 records why the production path uses shared output preparation plus provider receipts and rejects the non-convergent predictive cache-identity allowlist.

Supersedes #472.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -146,8 +146,17 @@ older closed tool-call/result unit while retaining a token-priced recent tail.
 Open and parallel tool units are indivisible, and a lone user payload is never
 eligible for the intra-turn fallback.
 
+A **Cache-Aligned Compaction Request** reuses the normal provider request's
+cacheable tool prefix and anchors retained history before the one-off
+summarization instruction. Provider-specific cache markers and TTL resolution
+come from shared request-preparation functions; cache-read usage is the source
+of truth for whether reuse occurred. Compaction never predicts cache identity
+from a hand-maintained list of provider options or environment variables.
+
 See [ADR 0005](docs/adr/0005-pairing-balanced-compaction-ranges.md) for the
 range-selection and retention trade-offs.
+See [ADR 0006](docs/adr/0006-align-compaction-with-provider-cache-prefixes.md)
+for the prompt-cache boundary and rejected predictive replay design.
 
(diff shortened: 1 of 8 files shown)


danny-avila:
@codex review

Please review the current PR head 40a8e2803fc06671e7e5dbf18a7c0cf0eb725e4a. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

danny-avila:
Performance results on exact head 40a8e2803fc06671e7e5dbf18a7c0cf0eb725e4a:

- Local hot-path CPU check, 100 tools × 5,000 preparations: direct median 2.559 ms; shared projection 2.554 ms (-0.2%, effectively noise).
- Live Anthropic before/after: prime created 6,408 cache tokens; the former unmarked compaction baseline read 0 cached tokens at 830 ms; the aligned projection read all 6,408 cached tokens at 773 ms.
- That is a 100% cache reuse conversion for the measured stable tool prefix and 57 ms / 6.9% lower request latency in this run. The token-reuse assertion, the full Anthropic summarization suite (9 passed), and the Bedrock suite all passed.

PR #479, opened 2026-08-29
🔁 fix: Stop Empty Summaries From Looping Compaction

danny-avila:
## Summary

I fixed the summarization recursion loop reported in [LibreChat#15294](https://github.com/danny-avila/LibreChat/issues/15294), where a run emits `{type: "summary", content: []}` over and over until the graph hits its recursion cap. The loop originates here, not in LibreChat — nothing on that side re-enters summarization.

An empty summarizer response compacts nothing, so the state the pruner sees on the next pass is identical to the one that just triggered. The summarize node reset the dedupe guard to `0` on that path (`markSummarizationTriggered(0)`), which made `shouldSkipSummarization` answer `false` for *every* later message count — the guard is written as `_lastSummarizationMsgCount > 0 && …`. The agent node then re-triggered on that unchanged state, routed back into the summarize node, got another empty response, and reset the guard again. Every pass billed a full summarizer call over the whole history and persisted a content-less summary block, until LangGraph threw `GraphRecursionError`.

- Recorded the current message count on the empty-output path instead of resetting the guard to `0`, so the identical state stops re-triggering.
- Added a consecutive-failure tally to `AgentContext` (`recordSummarizationFailure`, `summarizationFailures`, `summarizationExhausted`, capped at 3), which bounds the retries once new messages legitimately lift the count guard.
- Returned from the summarize node before `dispatchRunStep` once the tally is exhausted, so no further empty summary blocks are produced, persisted, or billed.
- Counted a provider failure the run declined to paper over with a metadata stub (overflow recovery and intra-turn compaction, where history is preserved instead) toward the same tally — that attempt made no progress either.
- Cleared the tally on a successful `setSummary` and on `reset()`, so a transient empty response costs nothing beyond the run it happened in.
- Short-circuited the trigger evaluation in `StandardGraph` on `summarizationExhausted`, and surfaced the tally in the existing skip log.

## Provenance

`markSummarizationTriggered(0)` was introduced by `ca586de` (squash-merged as #72, *Staged Context Compaction Pipeline with LLM Summarization*, 2026-03-20), whose commit message states the intent plainly:

> Also fixes the empty-output early return path: resets the summarization baseline via `markSummarizationTriggered(0)` so the next overflow can retry compaction instead of being permanently blocked.

The intent was right — an empty summary should not permanently block later compaction. The mechanism was the problem: because `shouldSkipSummarization` reads `_lastSummarizationMsgCount > 0 && …`, `0` is a sentinel meaning "no baseline recorded", which doesn't express "retry when the situation changes" but "never skip again, ever". So instead of *retry later*, it produced *retry immediately, forever, on byte-identical state*.

This fix preserves the original goal. Marking the current count still lets a genuinely changed conversation re-trigger, because a growing message count clears the guard on its own; the failure tally is what stops an unchanged-but-broken summarizer from spending calls indefinitely.

Two existing tests in `AgentContext.test.ts` (`shouldSkipSummarization — re-trigger after summary`) assert the `0`-means-unblock semantics directly, which is a fair part of why this survived review — the sentinel behavior read as intentional. Those tests still pass: they exercise `rebuildTokenMapAfterSummarization({})`, whose `0` write is immediately corrected by `markSummarizationTriggered(messagesToRetain.length)` on the success path.

**Affected releases:** every release from **v3.1.63** (2026-03-31) through **v3.7.7** — 113 tagged versions. LibreChat pins `^3.7.7` on `main` and `^3.2.46` on `v0.8.7`; both ranges resolve only to affected versions.

## Why it fires in practice

`extractResponseText` deliberately skips `thinking` / `reasoning_content` / `redacted_thinking` blocks and collects only `type: 'text'`, so any summarizer turn that ends without a text block yields `''`. Two routes get there, both reachable since #72:

1. **Reasoning-only response.** With no `summarization:` block configured, the summarizer self-summarizes: `buildSummarizationClientConfig` spreads `agentContext.clientOptions` wholesale, so the agent's `thinking` config carries into the summarizer. A turn that spends its output allowance on thinking returns thinking blocks and no text. This matches the reporter's setup exactly — Anthropic, `claude-fable-5`, no `librechat.yaml` summarization config.
2. **Tool call instead of text.** The summarizer is bound with the agent's tools, so a response of `[tool_use]` (or `[thinking, tool_use]`) also extracts to `''`, despite the prompt's "Only the checkpoint" instruction.

Neither route was covered by the metadata-stub fallback, which only fires when the call *throws*.

Worth noting as a **follow-up I deliberately did not fold in**: route 1 could be cut off at the source by not inheriting `thinking` into the self-summarize client options — the summary is text-only by construction, so a thinking budget there is pure cost. I left it out because changing thinking parameters alters the Anthropic request shape and would likely invalidate the cache prefix that #473 just aligned. That trade deserves its own PR and your call, not a quiet rider on a loop fix.

## Change Type

## Testing

Added a regression spec that drives a real `Run` through `processStream` with a `FakeListChatModel` split by model name: the agent model answers normally while the dedicated summarizer returns `''`. Reverting only the `markSummarizationTriggered` line makes it fail with `Recursion limit of 25 reached without hitting a stop condition` — the same failure mode as the report. With the fix the run completes after a single summarization attempt.

Also added four unit tests against `createSummarizeNode` covering the dedupe guard on empty output, the failure cap halting model calls after three consecutive empties, the tally clearing on a successful summary, and a preserved-history stub failure counting toward the cap.

### **Test Configuration**:

No API keys required — every new test uses `FakeListChatModel`.

```
npx jest src/summarization src/specs/summarization.test.ts src/agents/__tests__
npx tsc --noEmit
npx eslint src/
npm run build
```

Full suite: 4746 passed. The 4 failing suites are pre-existing environment gaps unrelated to this change — three live provider specs (`llm/anthropic`, `llm/google`, `llm/vertexai`) need credentials, and `durability-checkpoint.integration.test.ts` fails identically on a clean `main` because `mongodb-memory-server` cannot download a MongoDB binary in this sandbox.

## Checklist

https://claude.ai/code/session_015qnA9kToNcuJhX8VH4GBsH

### DIFF
--- a/src/agents/AgentContext.ts
+++ b/src/agents/AgentContext.ts
@@ -75,6 +75,14 @@ type ProgrammaticToolInstructionTarget = {
 };
 
 /**
+ * Consecutive summarization attempts that may return no usable summary before
+ * the run stops asking. Every such attempt spends a full model call over the
+ * whole history and leaves the message set exactly as it was, so the cap
+ * bounds spend as much as it bounds the compaction loop.
+ */
+const MAX_SUMMARIZATION_FAILURES = 3;
+
+/**
  * Encapsulates agent-specific state that can vary between agents in a multi-agent system
  */
 export class AgentContext {
@@ -376,6 +384,13 @@ export class AgentContext {
    * Summarization is allowed to fire again only when new messages appear.
    */
   private _lastSummarizationMsgCount: number = 0;
+  /**
+   * Consecutive summarization attempts that produced no usable summary.
+   * An empty or failed summary leaves the message set exactly as it was, so
+   * the next prune cycle would ask again on identical state. Cleared by
+   * `setSummary` and by `reset()`.
+   */
+  private _summarizationFailures: number = 0;
   /**
    * Forced compactions performed after a provider rejected a prompt as too
    * large. Bounds the recovery loop so a model that keeps refusing cannot
@@ -1209,6 +1224,7 @@ export class AgentContext {
     this.summaryTokenCount = this._durableSummaryTokenCount;
     this.summaryPrecedesMessages = this.durableSummaryPrecedesMessages;
     this._lastSummarizationMsgCount = 0;
+    this._summarizationFailures = 0;
     this.lastCallUsage = undefined;
     this.totalTokensFresh = false;
     this.restoreContextBudgetAfterOverflow();
@@ -1478,6 +1494,7 @@ export class AgentContext {
     this._durableSummaryTokenCount = tokenCount;
     this.durableSummaryPrecedesMessages = this.summaryPrecedesMessages;
     this._summaryVersion += 1;
+    this._summarizationFailures = 0;
     this.systemRunnableStale = true;
     this.pruneMessages = undefined;
   }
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

Please review the current PR head d3ef5ab. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

---

PR #329, opened 2026-07-23
fix: sanitize CodeAPI execution errors

upman:
## Summary

- Normalize CodeAPI transport and platform failures before they become tool output.
- Preserve only safe recovery signals, including bounded rate-limit delays and tenant execution stderr.
- Cover direct, bash, and programmatic execution paths.

## Why

CodeAPI failures currently include raw internal endpoints and response bodies in agent tool errors, which can expose deployment details in conversations and langfuse traces. Since we will expose langfuse traces to tenants, we need to clean it up.

## Validation

- `npm test -- --runInBand src/tools/__tests__/CodeApiAuthHeaders.test.ts`
- `npx tsc --noEmit`
- `npm run build`

PR #489, opened 2026-08-31
🧩 fix: Route Tool-Free Programmatic Calls to Plain Exec

### LINKED ISSUE #488: Two separate code-execution tools (execute_code / run_tools_with_bash) force the model to route, and picking wrong fails opaquely
`execute_code` (`CodeExecutor.ts`) and `run_tools_with_bash` (`BashProgrammaticToolCalling.ts`) are two entry points for what is conceptually one action — run code in the sandbox — differing only in whether tool stubs get injected. The model has to decide up front which one applies, and nothing in the tool descriptions or the framework prevents/corrects a wrong choice before it costs a round trip.

Concrete failure modes hit in production (local models, but nothing here is model-specific):

1. The model calls `run_tools_with_bash` for plain shell work that references no tool at all (`ls`, `find`, `python3 script.py`). Since no tool is referenced, `tools: []` goes to `/exec/programmatic`, which 400s. The client maps *any* 400/422 to the generic `CODE_API_INVALID_REQUEST_ERROR_MESSAGE` ("The code execution request was rejected. Please check the tool input and try again.") — `CodeExecutor.ts:135`, thrown from `ProgrammaticToolCalling.ts:526`. Nothing in that message hints at the actual cause (empty tool manifest), so the model retries the identical call repeatedly instead of switching tools.
2. Even after explicit system-prompt instructions telling the model which tool to use when, it regressed on a structurally identical case (jq-filtering an already-saved file — no tool call, but still routed through `run_tools_with_bash`). The boundary isn't enforced anywhere in the framework, only in prose the model has to re-derive per situation.

Suggestion: either (a) merge into one tool where the runtime decides whether tool stubs are relevant (e.g. based on whether the code references any registered stub, mirroring `filterBashToolsByUsage`'s own detection logic, which already exists for a different purpose), or (b) keep both but reject an empty-manifest `run_tools_with_bash` call client-side with a specific, actionable error ("no tools referenced — use execute_code instead") rather than sending it to the sandbox API and surfacing a generic rejection.

danny-avila:
Fixes #488

## Problem

`run_tools_with_bash` / `run_tools_with_code` had **no valid call** for code that references no tools.

- Omit `tool_manifest` → `selectProgrammaticTools` throws *"requires a tool_manifest when direct-only tools are configured"*.
- Send the honest answer, `[]` → codeapi's `/exec/programmatic` rejects it (`tools` must be a non-empty array), and `buildCodeApiHttpErrorMessage` maps that 400 to the generic `CODE_API_INVALID_REQUEST_ERROR_MESSAGE`. Nothing in that message names the cause, so the model retries the identical call instead of switching tools.

That genericization is deliberate (cc362b2b, #329), so the fix belongs on the client, not in un-sanitizing the upstream body.

**This is the default configuration, not an edge case.** `getAllowedCallers` falls back to `['direct']`, and hosts register action and built-in tools that way unconditionally — so an agent with both a code tool and any ordinary tool (exactly what #488 describes) always has a non-empty `directOnlyTools`, and could not run `ls` through the programmatic caller at all.

## Fix

Resolve the manifest in the runtime instead of making the model guess it.

**Derive an omitted manifest from the code rather than rejecting it.** Derivation intersects with the caller's allowed set, so a direct-only tool still cannot be reached — the caller-facing invariant is unchanged, and detection can only ever *narrow*. A miss drops a stub and surfaces as `not defined` / `command not found` inside the sandbox, never as access to something the caller wasn't already permitted to use.

**Treat an empty manifest as meaningful, not an error.** Code that calls no tools needs no stubs and no replay round trips, so remote callers route it to plain `/exec`, where it behaves exactly like the equivalent `execute_code` — against every deployed Code API, with no server change and no version skew. Bash sends raw code on that path: the `: &\nwait "$!"` guard exists for the programmatic replay wrapper (`preamble-bash.ts`), which `/exec` never applies.

Identifier normalization and usage extraction move to `toolIdentifiers.ts`, a leaf module both runtimes share, so the policy can derive without importing its own callers. `normalizeToPythonIdentifier`, `normalizeToBashIdentifier`, `extractUsedToolNames` and `extractUsedBashToolNames` are re-exported from their previous homes, so the package's public API is unchanged.

## Behavior

| call | before | after |
|---|---|---|
| `tool_manifest: []`, tool-free code | 400 → opaque "request was rejected" | runs via `/exec` |
| manifest omitted, direct-only tools present, tool-free code | throws "requires a tool_manifest" | runs via `/exec` |
| manifest omitted, direct-only tools present, code calls `get_weather` | throws "requires a tool_manifest" | `/exec/programmatic` with just `get_weather` |
| manifest names a direct-only tool | rejected | rejected (unchanged) |
| manifest omitted, no direct-only tools | full manifest sent | full manifest sent (unchanged) |

## Notes

- `/exec` ignores `timeout` (it destructures `{ user_id, lang, code, files }`), so a model-supplied timeout drops on the plain path. That matches `execute_code`, which has no timeout param at all, so it is not a regression against the tool that should have been called — honoring it is a separate one-line codeapi change.
- Cloudflare and local backends already tolerated an empty tool set and build a program with zero stubs; they get the derivation but need no routing change.
- Scope note for the issue: there are four sandbox entry points, not two — `execute_code`, `bash_tool`, `run_tools_with_code`, `run_tools_with_bash`. This makes the routing choice between them stop mattering for the failing case without merging them.

## Verification

- `npx tsc --noEmit` clean
- `npx eslint src/tools/` — 0 errors, 16 warnings (identical to `main`)
- `npx jest src/tools/` — 1550 passed, 59 suites
- Full suite: only `src/llm/{anthropic,google,vertexai}/llm.spec.ts` fail, identically on `main` (live provider credentials / ESM)
- `npm run check:circular-deps` — no circular dependencies
- 12 new tests in `ProgrammaticEmptyManifest.test.ts` covering both routes, the derivation safety property, session/file/runtime-hint propagation, and the unchanged paths

### DIFF
--- a/src/tools/BashProgrammaticToolCalling.ts
+++ b/src/tools/BashProgrammaticToolCalling.ts
@@ -386,6 +386,7 @@ export function createBashProgrammaticToolCallingTool(
             baseUrl,
             lang: 'bash',
             code,
+            timeout,
             sessionId: session_id,
             files,
             runtimeSessionHint,
--- a/src/tools/BashProgrammaticToolCalling.ts
+++ b/src/tools/BashProgrammaticToolCalling.ts
@@ -306,14 +306,14 @@ export function createBashProgrammaticToolCallingTool(
        * replays, and event-driven ToolNode configurations legitimately inject
        * an empty toolMap/toolDefs, so requiring them here would re-break the
        * tool-free call this route exists to serve. Injected context is still
-       * required to conclude that: with nothing injected at all, an empty
-       * selection means the host wired the tool up wrong, not that the code
-       * needs no tools. */
+       * required to conclude that — but an injected empty context counts:
+       * ToolNode legitimately yields no code-execution tools. Only a call with
+       * nothing injected at all is treated as a host wiring mistake. */
       const needsNoTools =
         effectiveTools.length === 0 &&
         (params.tool_manifest?.length === 0 ||
-          (toolDefs?.length ?? 0) > 0 ||
-          (disallowedToolDefs?.length ?? 0) > 0);
+          toolDefs != null ||
+          disallowedToolDefs != null);
 
       if (!needsNoTools) {
         if (toolMap == null || toolMap.size === 0) {
--- a/src/tools/BashProgrammaticToolCalling.ts
+++ b/src/tools/BashProgrammaticToolCalling.ts
@@ -302,22 +302,37 @@ export function createBashProgrammaticToolCallingTool(
         programmaticToolName,
       });
 
-      if (toolMap == null || toolMap.size === 0) {
-        throw new Error(
-          'No toolMap provided. ' +
-            'ToolNode should inject this from AgentContext when invoked through the graph.'
-        );
-      }
+      /* These guard the replay path only. A call that selected no tools never
+       * replays, and event-driven ToolNode configurations legitimately inject
+       * an empty toolMap/toolDefs, so requiring them here would re-break the
+       * tool-free call this route exists to serve. Injected context is still
+       * required to conclude that: with nothing injected at all, an empty
+       * selection means the host wired the tool up wrong, not that the code
+       * needs no tools. */
+      const needsNoTools =
+        effectiveTools.length === 0 &&
+        (params.tool_manifest?.length === 0 ||
+          (toolDefs?.length ?? 0) > 0 ||
+          (disallowedToolDefs?.length ?? 0) > 0);
 
-      if (toolDefs == null || toolDefs.length === 0) {
-        throw new Error(
-          'No tool definitions provided. ' +
-            'Either pass tools in the input or ensure ToolNode injects toolDefs.'
-        );
+      if (!needsNoTools) {
+        if (toolMap == null || toolMap.size === 0) {
+          throw new Error(
+            'No toolMap provided. ' +
+              'ToolNode should inject this from AgentContext when invoked through the graph.'
+          );
+        }
+
+        if (toolDefs == null || toolDefs.length === 0) {
+          throw new Error(
+            'No tool definitions provided. ' +
+              'Either pass tools in the input or ensure ToolNode injects toolDefs.'
+          );
+        }
       }
 
       const effectiveToolMap = projectProgrammaticToolMap(
-        toolMap,
+        toolMap ?? new Map(),
         effectiveTools
       );
 
@@ -332,7 +347,7 @@ export function createBashProgrammaticToolCallingTool(
           // eslint-disable-next-line no-console
           console.log(
             `[BashPTC Debug] Sending ${effectiveTools.length} tools to API ` +
-              `(selected from ${toolDefs.length})`
+              `(selected from ${toolDefs?.length ?? 0})`
           );
         }
 
@@ -366,7 +381,7 @@ export function createBashProgrammaticToolCallingTool(
 
         /* Raw `code`, not `preparedCode`: the `$!` guard exists for the
          * programmatic replay wrapper, which plain `/exec` never applies. */
-        if (effectiveTools.length === 0) {
+        if (needsNoTools) {
           return await runPlainExecution({
             baseUrl,
             lang: 'bash',
--- a/src/tools/BashProgrammaticToolCalling.ts
+++ b/src/tools/BashProgrammaticToolCalling.ts
@@ -14,11 +14,6 @@ import {
   selectRuntimeSessionHint,
 } from './CodeExecutor';
 import {
-  makeRequest,
-  executeTools,
-  formatCompletedResponse,
-} from './ProgrammaticToolCalling';
-import {
   projectProgrammaticToolMap,
   resolveProgrammaticToolDefinitions,
   selectProgrammaticTools,
@@ -29,10 +24,25 @@ import {
   createCodeApiRunTimeoutSchema,
   resolveCodeApiRunTimeoutMs,
 } from './ptcTimeout';
+import {
+  makeRequest,
+  executeTools,
+  runPlainExecution,
+  formatCompletedResponse,
+} from './ProgrammaticToolCalling';
+import {
+  extractUsedBashToolNames,
+  normalizeToBashIdentifier,
+} from './toolIdentifiers';
 import { INTENT_PROPERTY } from '@/tools/intentArg';
 import { Constants } from '@/common';
 
 config();
+
+export {
+  extractUsedBashToolNames,
+  normalizeToBashIdentifier,
+} from './toolIdentifiers';
 
 // ============================================================================
 // Constants
@@ -41,33 +51,6 @@ config();
 const DEFAULT_MAX_ROUND_TRIPS = 20;
 const DEFAULT_RUN_TIMEOUT_MS = resolveCodeApiRunTimeoutMs();
 const BASH_LAST_BACKGROUND_PID_GUARD = ': &\nwait "$!"';
-
-/** Bash reserved words that get `_tool` suffix when used as function names */
-const BASH_RESERVED = new Set([
-  'if',
-  'then',
-  'else',
-  'elif',
-  'fi',
-  'case',
-  'esac',
-  'for',
-  'while',
-  'until',
-  'do',
-  'done',
-  'in',
-  'function',
-  'select',
-  'time',
-  'coproc',
-  'declare',
-  'typeset',
-  'local',
-  'readonly',
-  'export',
-  'unset',
-]);
 
 // ============================================================================
 // Description Components
@@ -116,8 +99,9 @@ ${EXAMPLES}
 ${CORE_RULES}`;
 
 const TOOL_MANIFEST_DESCRIPTION =
-  'Exact registered tool names used by the code. Required when direct-only tools are configured; ' +
-  'validated before execution starts.';
+  'Exact registered tool names used by the code, validated before execution starts. ' +
+  'Omit it to have the runtime derive the manifest from the code; pass [] when the code ' +
+  'calls no tools at all.';
 
 // ============================================================================
 // Schema
@@ -217,51 +201,6 @@ export function normalizeBashToolResultsForReplay(
 // ============================================================================
 
 /**
- * Normalizes a tool name to a valid bash function identifier.
- * 1. Replace hyphens, spaces, dots with underscores
- * 2. Remove any other invalid characters
- * 3. Prefix with underscore if starts with number
- * 4. Append `_tool` if it's a bash reserved word
- */
-export function normalizeToBashIdentifier(name: string): string {
-  let normalized = name.replace(/[-\s.]/g, '_');
-  normalized = normalized.replace(/[^a-zA-Z0-9_]/g, '');
-
-  if (/^[0-9]/.test(normalized)) {
-    normalized = '_' + normalized;
-  }
-
-  if (BASH_RESERVED.has(normalized)) {
-    normalized = normalized + '_tool';
-  }
-
-  return normalized;
-}
-
-/**
- * Extracts tool names that are actually called in the bash code.
- * Bash functions are invoked as commands (no parentheses), so we match
- * the normalized name as a word boundary.
- */
-export function extractUsedBashToolNames(
-  code: string,
-  toolNameMap: Map<string, string>
-): Set<string> {
-  const usedTools = new Set<string>();
-
-  for (const [bashName, originalName] of toolNameMap) {
-    const escapedName = bashName.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
-    const pattern = new RegExp(`\\b${escapedName}\\b`, 'g');
-
-    if (pattern.test(code)) {
-      usedTools.add(originalName);
-    }
-  }
-
-  return usedTools;
-}
-
-/**
  * Filters tool definitions to only include tools actually used in the bash code.
  */
 export function filterBashToolsByUsage(
@@ -355,6 +294,8 @@ export function createBashProgrammaticToolCallingTool(
           ? toolCall.name
           : Constants.BASH_PROGRAMMATIC_TOOL_CALLING);
       const effectiveTools = selectProgrammaticTools({
+        code,
+        runtime: 'bash',
         requestedToolNames: params.tool_manifest,
         allowedToolDefs: toolDefs,
         disallowedToolDefs,
@@ -422,6 +363,22 @@ export function createBashProgrammaticToolCallingTool(
           selectedRuntimeSessionHint !== ''
             ? selectedRuntimeSessionHint
             : undefined;
+
+        /* Raw `code`, not `preparedCode`: the `$!` guard exists for the
+         * programmatic replay wrapper, which plain `/exec` never applies. */
+        if (effectiveTools.length === 0) {
+          return await runPlainExecution({
+            baseUrl,
+            lang: 'bash',
+            code,
+            sessionId: session_id,
+            files,
+            runtimeSessionHint,
+            proxy,
+            authHeaders: initParams.authHeaders,
+            executionProfile: initParams.executionProfile,
+          });
+        }
 
         let response = await makeRequest(
           EXEC_ENDPOINT,
--- a/src/tools/ProgrammaticCallerPolicy.ts
+++ b/src/tools/ProgrammaticCallerPolicy.ts
@@ -70,10 +70,25 @@ export function selectProgrammaticTools(args: {
   const requestedToolNames = args.requestedToolNames;
 
   if (requestedToolNames == null) {
-    if (disallowedToolDefs.length > 0) {
-      return deriveReferencedToolDefs(allowedToolDefs, args.code, args.runtime);
+    if (disallowedToolDefs.length === 0) {
+      return allowedToolDefs;
     }
-    return allowedToolDefs;
+
+    /* Derive over the disallowed set as well. A direct-only tool the code
+     * references must still be rejected: some runtimes define native tool
+     * functions unconditionally, so treating an unmatched name as absent would
+     * let the call reach one the caller may not use. */
+    const referencedDisallowed = deriveReferencedToolDefs(
+      disallowedToolDefs,
+      args.code,
+      args.runtime
+    );
+    assertDisallowedToolUsage(
+      new Set(referencedDisallowed.map((toolDef) => toolDef.name)),
+      args.programmaticToolName
+    );
+
+    return deriveReferencedToolDefs(allowedToolDefs, args.code, args.runtime);
   }
 
   const disallowedNames = new Set(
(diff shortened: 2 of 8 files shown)


danny-avila:
@codex review

danny-avila:
All four addressed in 0f99c6ab — each verified against the source before changing anything.

**P1 (native tool bypass) — confirmed, real regression.** `createPythonProgram` prepends `createPythonNativeToolSource` unconditionally and only gates the *aliases* on `toolDefs`, so every native function is a global regardless of selection. Before this PR the omitted-manifest path threw, which is what kept that unreachable; deriving turned the throw into an empty selection and let the code run. Fixed by deriving over `disallowedToolDefs` as well and routing hits through the existing `assertDisallowedToolUsage`, so a derived reference to a direct-only tool is rejected exactly like an explicit one.

Worth noting separately: the unconditional native globals are a pre-existing hole independent of this PR — `tool_manifest: ['calculator']` plus `write_file(...)` reaches it today. Only defining selected native functions is the real fix; that belongs in its own change.

**P2 (guard ordering) — confirmed.** `getProgrammaticTools()` builds `toolMap`/`toolDefs` from `codeExecutionTools` and skips non-direct tools in `eventDrivenMode`, so both can legitimately be empty while `disallowedToolDefs` is populated. Guards now apply only when tools are actually needed.

One refinement over the suggestion: an empty selection is not on its own sufficient to conclude the code needs no tools. With *nothing* injected — no allowed defs, no disallowed defs, no explicit manifest — an empty selection means the host wired the tool up wrong, and silently running the code would replace `No toolMap provided` with a confusing `command not found`. The route requires an explicit empty manifest, or some injected context to have derived against. The three existing `throws error when no toolMap/toolDefs` tests still pass unchanged, which is what caught this.

**P2 (async wrapper) — confirmed.** `createProgrammaticPayload` wraps Python in `async def __user_main__()` (`preamble.ts:796`); plain `/exec` does not. The plain path now mirrors that wrapper. It is self-contained, so a later divergence in the Code API's copy can't break this route.

**P2 (identifier collisions) — confirmed, and it mattered more than stated.** The Code API *rejects* bash collisions server-side via `findBashToolNameCollision`, so before this PR a collision failed loudly once both tools were sent; deriving one of the pair would have converted that into silent wrong-tool dispatch. Rather than requiring an explicit manifest — which carries the same ambiguity for Python, where there is no server-side guard — every colliding definition is kept, so behavior is identical to sending both and the server's loud rejection still fires.

7 new tests cover the four paths. `tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1557 passing, no circular deps.

danny-avila:
@codex review

danny-avila:
Round 2 addressed in 84d29b2c.

**P1 (indirect native access) — confirmed, and it changed my approach.** You're right that the regex only covers literal `write_file(...)`; `globals()["write_file"]` sails past it. Chasing that with better detection is the wrong move — the regex should never have been the boundary.

So derivation no longer substitutes for an explicit manifest on the path where the runtime doesn't gate its own stubs. `runtimeExposesUnselectedTools` keeps the original rejection for Cloudflare Python; everywhere else (bash in all three backends, remote Python, local) stubs are generated from the selected set, so absence enforces the boundary and derivation is safe.

To be explicit about what this does *not* fix: the unconditional native definitions are a pre-existing hole that an explicit manifest doesn't close either — `tool_manifest: ['calculator']` plus `write_file(...)` reaches it on `main` today. Only defining selected native functions is the real fix, and it needs its own change and review; deleting names from `globals()` after the fact risks breaking native tools that call each other. What matters here is that this PR no longer widens it.

**P2 (comments and strings) — confirmed.** `stripNonCodeText` blanks line comments and string literals (including triple-quoted Python) before matching, so `print("write_file(")` and `echo write_file` no longer trigger a rejection. I did not go as far as command-position parsing for bash: this only ever reduces false positives, and with the P1 change above the boundary no longer rests on it.

**P2 (local guards) — confirmed, and I'd missed it.** I scoped the guards in both remote factories and left the identical pair in `runLocalProgrammaticTool` untouched. Same fix applied, including the injected-context requirement so a genuinely unconfigured runner still fails loudly rather than reporting `command not found`.

6 more tests. `tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1563 passing, no circular deps. Previous round's CI was 13/13.

danny-avila:
@codex review

danny-avila:
Round 3 addressed in 600de152.

**P2 (interpolation) — confirmed, and it was a real false negative.** Blanking string literals erased executable code along with prose: `print(f"{await calculator(1)}")` and `echo "$(calculator '{}')"` both lost the call, so derivation would have routed them to plain `/exec` with no stub and failed at runtime. Worse than the false positive I was fixing.

F-prefixed Python literals now survive the pass, and bash erases only single-quoted text — double quotes interpolate, so they have to stay visible. `b"calculator("` and `# calculator` are still blanked. Four tests cover both directions.

**P2 (timeout) — confirmed, with one correction on severity.** The value was dropped entirely; it now rides the plain-path body, so the cap applies the moment the Code API honors it on `/exec`.

Where I'd push back: this isn't unbounded. `/exec` reads only `user_id`, `lang`, `code` and `files`, but the worker applies its own `timeouts.run` to every job, so a runaway script is still bounded by the sandbox's limit — what's lost is a *lower* model-supplied cap, not all enforcement. Honoring `timeout` on `/exec` is a one-line codeapi change and the right permanent fix; forwarding the field here means no client change is needed when it lands. I'd rather do that than abort the fetch client-side, which would orphan the sandbox run and swap a clean server error for a generic one.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1568 passing, no circular deps. Round 2 CI was 13/13.

danny-avila:
@codex review

danny-avila:
Round 4 in 78278aa6 — two taken, one declined.

**P2 (f-string literal text) — confirmed, my round-3 fix overcorrected.** Keeping the whole f-string made prose executable, so `f"example: write_file(path)"` was rejected as a direct-only call. Now only `{...}` replacement fields survive; the literal portions are blanked, and `{{`/`}}` are treated as text. Three tests pin both directions.

**P2 (collisions) — confirmed, and the new evidence changed my mind.** You're right that retaining both doesn't disambiguate: the local Python wrapper emits two `async def report_tool(...)` and the later one wins. "Keep both" only made the ambiguity explicit at codegen instead of resolving it.

I went further than the suggestion in one way. An *explicit* manifest naming both collapses identically, so rejecting only the derived path would leave the same silent mis-dispatch reachable through the documented route. The check now runs on the final selection however it was produced — derived or explicit — with a message naming both tools and the identifier they share. A manifest naming just one of the pair is still fine, since codegen then emits only that stub.

**P2 (timeout) — declining, with reasoning.** The proposed remedy is "an actual supported cancellation/limit mechanism or must not be used for these calls." The second half would reintroduce the 400 this PR exists to fix for any call that sets `timeout`, which is a worse outcome than the gap it closes.

On the first half: aborting the fetch client-side would return within the requested bound, but the sandbox job keeps running server-side — it can still complete and register files under `/mnt/data` while the model has been told it timed out. Trading a clean server-enforced error for a client-side lie about what happened isn't an improvement.

The factual correction from my last comment stands and is load-bearing here: this is not unbounded. The worker applies `timeouts.run` to every job including `/exec`, so a runaway script is capped — what's lost is a *lower* model-supplied value. The real fix is honoring `timeout` in codeapi's `/exec` handler, which today reads only `user_id`, `lang`, `code` and `files`; the field is already forwarded, so that lands with no client change. I'd rather carry a documented, bounded gap than paper over it.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1575 passing, no circular deps. Round 3 CI was 13/13.

danny-avila:
@codex review

danny-avila:
Round 5 in the latest commit — confirmed, and it was my own regex.

`\{[^{}]*\}` stops at the first inner brace, so `f"{await calculator(payload={'x': 1})}"` preserved only the dict literal and blanked the surrounding call — the same false-negative class as round 3, resurfacing for nested braces. Replaced with a balanced-brace scan.

One deliberate choice in it: an unterminated field is preserved rather than blanked. The two failure directions are not symmetric — keeping too much costs a loud, actionable rejection naming the tool, while dropping a real call routes to plain `/exec` and fails at runtime with an undefined stub. Four tests cover nested dicts, a field beside another field, the unterminated case, and doubled braces still reading as prose.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1579 passing, no circular deps. Round 4 CI was 13/13.

danny-avila:
@codex review

danny-avila:
Round 6 in fe60a40e — all three, and the third one let me delete the mitigation I'd been defending.

**P2 (Cloudflare Python rejects tool-free code) — confirmed, and you were right the first time.** I'd declined "expose only selected native functions" in round 2 as scope creep with a risk of breaking natives that call each other. I checked that assumption instead of repeating it: the nine native tools call only their own `_`-prefixed helpers (`_run`, `_format_run`, `_line_window`, `_resolve`, …) and never one another by public name. So withdrawing the unselected names from `globals()` after the aliases bind is safe.

That closes the pre-existing hole an explicit manifest never closed, and with the runtime enforcing the boundary itself, `runtimeExposesUnselectedTools` is deleted — Cloudflare Python now derives tool-free code like everywhere else. Better outcome than the blunt flag, and it removes a concept from the diff.

**P2 (quoted braces) — confirmed.** The scanner counted braces inside quoted text within a replacement field, so `f"{'}' + await calculator()}"` closed early and blanked the call. It now tracks quote state and backslash escapes.

**P2 (bash `#` inside double quotes) — confirmed.** `(?:^|\s)#` fired inside `"..."`, erasing `echo "label # $(calculator '{}')"`. The bash stripper is a scanner now: single-quoted text is blanked, double-quoted content stays visible because it interpolates, and a comment opens only at a word start outside quotes. `calculator#1` is no longer read as a comment either.

Note both strippers moved from regex to scanners for the same reason — this class of finding is what you get approximating a lexer with a pattern, and the scanner form ends it rather than deferring the next case.

`tsc` clean, eslint 0 errors (16 warnings, same as `main` — one I'd added is gone), `npx jest src/tools/` 1584 passing, no circular deps. Round 5 CI was 13/13. One unrelated flake seen and re-confirmed green: `escalates to SIGKILL when timeoutMs elapses` in the local-execution suite is timing-sensitive.

danny-avila:
@codex review

danny-avila:
Round 7 in bb11c08e — one taken, one I'm answering rather than chasing.

**P2 (comments after control operators) — confirmed and fixed.** A comment opens at the start of a word, and `;`, `&`, `|`, `(`, `)` and backtick all end the previous one. `echo ok;# write_file is unavailable` no longer reads as a reference. Three tests, including that a real call after `;` still registers.

**P1 (`_run` survives the withdrawal) — correct, and it establishes that this boundary is not enforceable by name removal at all.** Not just `_run`: the generated module's first line is

```python
import asyncio, fnmatch, glob, json, os, pathlib, re, shlex, shutil, subprocess, sys, tempfile
```

and user code runs inside `__lc_user_main__()` in those same globals. So `subprocess.run(...)` is directly callable from any Cloudflare Python programmatic call — with a correct explicit manifest, on `main`, today, with or without this PR. Popping `_run` would leave `subprocess`; popping that leaves `os.system`, `__import__`, and so on. Each additional pop is theatre against a caller who is trying.

So I'm not adding more pops, and I'm not restoring the manifest requirement either — it provides no protection against this, only cost. What the withdrawal does buy is real and worth keeping: it stops *accidental* reach through the public tool name, which is what `allowed_callers` is actually expressing — operator configuration intent, not adversarial containment. I've recorded that boundary in the code rather than leaving the guarantee overstated.

The genuine fix is a restricted execution namespace for the generated program — running user code with an explicit `globals` dict instead of the module's, so the import surface is chosen rather than inherited. That's a change to how the program is generated, it needs its own design and review, and it is independent of this PR. Happy to open it as a separate issue.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1587 passing, no circular deps. Round 6 CI was 13/13, one job still running at time of writing.

danny-avila:
@codex review

danny-avila:
Round 8 in 1684ea24 — both taken.

**P2 (injected empty context) — confirmed, and your framing fixed my condition.** I was testing definition *length*, which conflates "ToolNode injected an empty context because nothing is nested" with "nothing was injected at all". Presence is the right test: `toolDefs != null || disallowedToolDefs != null`. An agent whose only tool is the programmatic caller now runs tool-free code, while a call with no injected context still fails loudly with `No toolMap provided` — the three manual-invocation tests that pin that behavior still pass untouched.

Applied to the local runner too, which carries the same condition.

**P2 (quoted command names) — confirmed.** Quote removal precedes command lookup, so `'calculator' '{}'` really does call `calculator`, and blanking every single-quoted word dropped it. The scanner now tracks command position: single quotes are blanked as arguments, where they are prose, and preserved in command position, where they are the command. `echo 'calculator'` is still prose; `echo ok; 'calculator' '{}'` is still a call.

That also makes the bash stripper principled rather than a pile of special cases — it now distinguishes command position from argument position, which is the actual rule the earlier findings were each approximating.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1591 passing, no circular deps. Round 7 CI was 13/13. The `escalates to SIGKILL when timeoutMs elapses` flake reappeared once (6463ms against a 6000ms bound) and passes in isolation — timing-sensitive, unrelated to this branch.

danny-avila:
@codex review

danny-avila:
Round 9 in the latest commit — confirmed.

Bash treats an assignment word as a prefix, not the command, so `FOO=bar 'calculator' '{}'` really does invoke `calculator`; the scanner held argument position across it and blanked the quoted name. Command position now resumes once the assignment value ends.

One thing worth flagging from implementing it: my first attempt blanked the assignment value wholesale, on the reasoning that a value is data rather than a command. That broke `raw=$(get_weather '{"city":"SF"}')` — which is the exact shape the tool description tells the model to use for capturing tool output, and it was caught by an existing test rather than a new one. The value is scanned normally now; only its quoted arguments are blanked, and the substitution inside stays visible.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1594 passing, no circular deps. Round 8 CI was 13/13.

danny-avila:
@codex review

danny-avila:
Round 10 in 777f469e — confirmed and fixed.

Requiring `\s*\(` missed any tool that is aliased or passed on: `fn = calculator; await fn(a=1)` selected nothing, so no stub was generated and the run broke. The derivation matcher now matches the bare word, the same way bash already did.

Worth noting this one predates the PR — the call-expression requirement is `extractUsedToolNames`'s existing behavior on `main`, which the derivation inherited. I've left that public function's semantics alone (it has its own tests pinning the `(` requirement) and changed only the private matcher.

The direction is the safe one: comments and literal text are blanked before matching, so what survives is code, and over-matching adds a stub the sandbox ignores rather than dropping one it needs.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1597 passing, no circular deps. Round 9 CI was 13/13.

danny-avila:
@codex review

danny-avila:
Round 11 in 18ade69b — both confirmed.

**P2 (literals inside replacement fields).** A quoted token in a field is data, so `f"{'write_file'}"` was being rejected as a disallowed reference. Quoted spans inside a field are blanked now while the expression around them survives — `f"{await write_file(path='a')}"` still resolves to the call, with the `'a'` argument dropped.

**P2 (bash control keywords).** `if`, `while`, `until`, `then`, `do`, `time`, `coproc` and `!` precede a command rather than being one, so they keep command position open. `if 'calculator' '{}'; then echo ok; fi` no longer blanks the quoted executable.

Implementation note worth recording: the first cut of the field-literal fix silently blanked from index 0 because the span start was never assigned — a `quoteStart` initialized to `-1`. The two existing nested-brace tests caught it immediately, which is the second time this session that a test written for an earlier round caught a regression in a later one.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/__tests__/ProgrammaticEmptyManifest.test.ts` 64 passing, no circular deps. Round 10 CI was 13/13.

The `escalates to SIGKILL when timeoutMs elapses` test failed again in a full-suite run (6306ms against a 6000ms bound) and passes both in isolation and on clean `main` — it is load-sensitive and unrelated to this branch, but it will keep surfacing in full runs on a busy machine.

danny-avila:
@codex review

danny-avila:
Closing in favor of #490, which ships the part of this that fixes #488 and drops the part that generated most of the review.

## Why

28 findings over 12 rounds, by root cause:

| Root cause | Findings | Converged? |
|---|---|---|
| Manifest derivation from source text | **13 (46%)** | **No** |
| Capability containment by name removal | **6 (21%)** | **No** |
| Guard/context placement | 3 | Yes, by r8 |
| Timeout contract | 2 | Yes, by r4 |
| Identifier collisions | 2 | Yes, by r4 |
| Plain-route contract | 2 | Yes |

68% of findings came from two categories, and they were the only two still producing new findings at round 12. A converging category is a bug list; a non-converging one is a design error.

The derivation promised "given source text, return the tools this code will call" — a claim about program semantics — and delivered 314 lines approximating two language lexers. The review walked that to its conclusion: round 10 said `\bname\s*\(` was too strict (missed `fn = calculator`), round 12 said `\bname\b` was too loose (`archive.read_file()`). The only thing between those positions is a real binding resolver.

Deleting it makes complexity vanish rather than move, because the information already exists at the caller: **the model wrote the code, so it knows which tools it called.** `tool_manifest` is it saying so. This PR re-derived from text a fact the caller had already been asked to declare.

The actual bug in #488 was never that the model can't declare its tools — it was that the runtime rejected the honest declaration, `[]`. #490 fixes that in 540 insertions against this PR's 1709, and keeps the parts that were bounded and provable: the plain-`/exec` route, the Python async wrapper, the guard scoping, and collision rejection.

Category B was downstream of the same decision — name removal was only needed to backstop derivation, and it isn't containment anyway. Filed as #491 with the paths this PR's approach couldn't close.

## Credit where it's due

Two of the P1s here were real regressions this branch introduced, and the round-7 and round-12 findings are what established that name removal can't enforce the boundary at all. The `_run` and `TOOLS`-table findings became #491.

Also correcting one of my own: finding #28 was a false positive I manufactured. `/exec` **does** honor `runtime_session_hint` (`router.ts:148`) — my code comment overstated a limitation that only ever applied to `timeout`, and the review reasoned from my comment. #490 states it accurately.

Follow-ups filed: #491 (Cloudflare sandbox containment), #492 (codeapi honoring `timeout` on `/exec`).

PR #490, opened 2026-09-01
🧩 fix: Run Tool-Free Programmatic Code Instead of Rejecting It

### LINKED ISSUE #488: Two separate code-execution tools (execute_code / run_tools_with_bash) force the model to route, and picking wrong fails opaquely

danny-avila:
Fixes #488. Replaces #489 — see "What this deliberately does not do" below for why.

## Problem

`run_tools_with_bash` / `run_tools_with_code` had no way for a model to say *"this code calls no tools."*

The honest answer, `tool_manifest: []`, is exactly what codeapi's `/exec/programmatic` rejects (`tools` must be a non-empty array, `programmatic-router.ts:482`). `buildCodeApiHttpErrorMessage` maps that 400 to the generic `CODE_API_INVALID_REQUEST_ERROR_MESSAGE`, which names nothing actionable — so the model retries the identical call rather than switching tools. That genericization is deliberate (cc362b2b, #329), so the fix belongs on the client.

## Fix

**An empty selection is meaningful, not an error.** Code that calls no tools needs no stubs and no replay round trips, so it routes to plain `/exec`, where it behaves exactly like the equivalent `execute_code` — against every deployed Code API, with no server change and no version skew.

- **Python keeps its wrapper.** `/exec/programmatic` wraps user code in `async def __user_main__()` (`preamble.ts:796`) so top-level `await` works; plain `/exec` does not. The plain path mirrors it, so valid programmatic input doesn't become a SyntaxError.
- **Bash sends raw code.** The `: &\nwait "$!"` guard exists for the programmatic replay wrapper (`preamble-bash.ts`), which `/exec` never applies.
- **Guards move behind the decision.** `toolMap`/`toolDefs` protect the replay path, and event-driven ToolNode configurations legitimately inject an empty context. Injected context is still *required* to conclude a call needs no tools, so a genuinely unconfigured runner still fails loudly with `No toolMap provided`.

**Ambiguous identifiers are rejected.** Two tools that normalize to the same runtime identifier are indistinguishable once stubs are emitted — the generated Python defines both and the later one wins. Failing beats dispatching by declaration order. Applies to explicit manifests and the all-tools fallback alike.

## What this deliberately does not do

Omitting `tool_manifest` when direct-only tools are configured still errors. That error is now *actionable* — `[]` is a valid answer and the message names the field — so the model self-corrects in one turn.

#489 tried to remove that turn by deriving the manifest from the submitted code. That's a claim about program semantics answerable only by parsing Python and Bash, and it produced 19 of that PR's 28 review findings in two categories that never converged — one round arguing the matcher was too strict, a later one arguing the same matcher was too loose. The model already knows which tools it called; it wrote the code. Asking it beats re-deriving it from text.

## Behavior

| call | before | after |
|---|---|---|
| `tool_manifest: []`, tool-free code | 400 → opaque "request was rejected" | runs via `/exec` |
| ToolNode injects an empty context | `No toolMap provided` | runs via `/exec` |
| nothing injected at all | `No toolMap provided` | unchanged |
| manifest omitted, direct-only tools present | `requires a tool_manifest` | unchanged, now actionable |
| manifest omitted, no direct-only tools | full manifest sent | unchanged |
| manifest names a direct-only tool | rejected | unchanged |
| two tools sharing one identifier | silent mis-dispatch | rejected |

## Notes

- `/exec` honors `session_id`, `files` and `runtime_session_hint`; it does not read `timeout` (`router.ts:136`). The field is forwarded so the cap applies as soon as codeapi honors it — filed as #492. Until then the worker's own `timeouts.run` bounds these calls, the same as `execute_code`.
- Cloudflare and local backends already tolerate an empty tool set and build a program with zero stubs; they need no routing change.
- Related pre-existing issue filed separately: #491.

## Verification

- `npx tsc --noEmit` clean
- `npx eslint src/tools/` — 0 errors, 16 warnings (identical to `main`)
- `npx jest src/tools/` — 1548 passing, 59 suites
- `npm run check:circular-deps` — none
- 13 new tests covering both routes, the python wrapper, session/file/hint/timeout propagation, collision rejection, and every unchanged path above

540 insertions against #489's 1709.

### DIFF
--- a/src/tools/BashProgrammaticToolCalling.ts
+++ b/src/tools/BashProgrammaticToolCalling.ts
@@ -14,6 +14,7 @@ import {
   selectRuntimeSessionHint,
 } from './CodeExecutor';
 import {
+  assertUnambiguousIdentifiers,
   projectProgrammaticToolMap,
   resolveProgrammaticToolDefinitions,
   selectProgrammaticTools,
@@ -356,7 +357,6 @@ export function createBashProgrammaticToolCallingTool(
           ? toolCall.name
           : Constants.BASH_PROGRAMMATIC_TOOL_CALLING);
       const effectiveTools = selectProgrammaticTools({
-        normalizeIdentifier: normalizeToBashIdentifier,
         requestedToolNames: params.tool_manifest,
         allowedToolDefs: toolDefs,
         disallowedToolDefs,
@@ -389,6 +389,12 @@ export function createBashProgrammaticToolCallingTool(
           );
         }
       }
+
+      assertUnambiguousIdentifiers(
+        effectiveTools,
+        normalizeToBashIdentifier,
+        programmaticToolName
+      );
 
       const effectiveToolMap = projectProgrammaticToolMap(
         toolMap ?? new Map(),
--- a/src/tools/BashProgrammaticToolCalling.ts
+++ b/src/tools/BashProgrammaticToolCalling.ts
@@ -14,11 +14,6 @@ import {
   selectRuntimeSessionHint,
 } from './CodeExecutor';
 import {
-  makeRequest,
-  executeTools,
-  formatCompletedResponse,
-} from './ProgrammaticToolCalling';
-import {
   projectProgrammaticToolMap,
   resolveProgrammaticToolDefinitions,
   selectProgrammaticTools,
@@ -29,6 +24,12 @@ import {
   createCodeApiRunTimeoutSchema,
   resolveCodeApiRunTimeoutMs,
 } from './ptcTimeout';
+import {
+  makeRequest,
+  executeTools,
+  runPlainExecution,
+  formatCompletedResponse,
+} from './ProgrammaticToolCalling';
 import { INTENT_PROPERTY } from '@/tools/intentArg';
 import { Constants } from '@/common';
 
@@ -117,7 +118,7 @@ ${CORE_RULES}`;
 
 const TOOL_MANIFEST_DESCRIPTION =
   'Exact registered tool names used by the code. Required when direct-only tools are configured; ' +
-  'validated before execution starts.';
+  'validated before execution starts. Pass [] when the code calls no tools at all.';
 
 // ============================================================================
 // Schema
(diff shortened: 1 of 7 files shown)


danny-avila:
@codex review

danny-avila:
Confirmed and fixed — and it found a seam I'd placed wrong.

`filterNativeTools` narrows the selection right after `selectProgrammaticTools` returns, so running the collision check inside selection judged definitions that Cloudflare never emits as stubs. A collision among those isn't a conflict.

The underlying mistake was putting the check in the policy at all. Selection answers "may this caller use these tools" — an authorization question that doesn't care how a name is spelled in a runtime. Identifier ambiguity is a property of *code generation*, which binds by identifier. So the assertion moved to each backend's effective set, and `selectProgrammaticTools` lost the `normalizeIdentifier` parameter it never should have carried.

Nice side effect: the policy is back to pure authorization, and the new test that a tool-free call with colliding definitions still runs (no stubs, no conflict) falls out naturally rather than needing a special case.

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1549 passing, no circular deps. Previous commit's CI was 13/13.

danny-avila:
@codex review

danny-avila:
Fair, and fixed in the latest commit.

The point lands harder than a style note: these are *routing* tests whose entire job is asserting what goes on the wire, and a permissive body type plus a permissive fetch mock means a payload could drift from the production contract while the tests kept passing — checking the wrong thing, greenly. The harness now names the three contracts it exercises: `SandboxRequestBody`, `ProgrammaticArgs`, and `InjectedContext`. `requests[0].body.tools` is typed, so the `as t.LCTool[]` assertion cast is gone too.

Three `as never` casts remain, on `.invoke()`. Those are deliberate: ToolNode injects `toolMap`/`toolDefs`/`session_id`/`_injected_files` onto the tool call at runtime, and LangChain's public `invoke` signature doesn't model those extras — the cast is at the boundary where the test plays ToolNode's role, with the payload itself now fully typed on the way in. Same pattern the existing suites use (`LocalExecutionTools.test.ts`, `hitl.test.ts`).

`tsc` clean, eslint 0 errors (16 warnings, same as `main`), `npx jest src/tools/` 1549 passing, no circular deps. Previous commit's CI was 13/13.

danny-avila:
@codex review

PR #496, opened 2026-09-02
⏭️ feat: Discard-and-Restart Preemption for Turns With Nothing to Keep

danny-avila:
## Summary

A preempt request could only ever be honored by **sealing**, and `canSealPreempt` requires non-empty *text* on the accumulated chunk. Two consequences followed, both in the window users reach for an interrupt most:

- a turn streaming only reasoning never becomes sealable, so the request waits out the whole turn;
- a provider that has sent nothing at all is never even polled — the only `shouldPreempt` read lives inside the chunk loop (`src/llm/invoke.ts`), so silence means no read.

Either way the queued message lands as a terminal continuation *after* the answer it was meant to redirect.

Both cases now convert to a **restart**: the provider request is torn down, its partial output discarded, the `PreemptBoundary` hook injects, and the model is called again with the injection appended to the same prompt.

## Why a discard is safe where a seal is not

A seal *keeps* the partial assistant turn, which is exactly why it needs non-empty text — the injected user turn must not be preceded by an empty assistant message. A discard leaves **no** assistant turn behind, so the injected turn sits directly after the previous user turn. Adjacent user turns are native on Anthropic, OpenAI and Gemini, and already normalized by `coalesceAdjacentUserTurns` for the `strictAlternationProviders` set.

## Design notes

**Seal always wins.** `resolvePreemptAction` prefers a seal wherever one is available, so a turn that has already produced an answer never loses it to a restart.

**The grace window is correctness, not politeness.** A restart converts only once the request outlives `DEFAULT_PREEMPT_RESTART_GRACE_MS` (2s, per-run overridable via `StreamPreemption.restartGraceMs`), for two independent reasons:

- reasoning usually precedes text by moments, and discarding a turn that was about to become sealable trades a kept answer for a re-issued request. The existing `preemptSeal.test.ts` case that steers a multi-item reasoning turn caught this directly — an eager first cut turned that seal into a discard;
- the chunk a restart is judged against is what the **consumer** has accumulated, and the stream buffers a chunk ahead of it. An empty accumulation therefore does not prove the provider produced nothing; it can also mean the first chunk is in flight. Converting on emptiness alone would discard that chunk — text included — on a race no caller can see. This is why the window applies to a silent provider too.

**The discardable shape is a whitelist.** Every content block must be a recognizable reasoning block; an unrecognized block refuses. Blacklisting would silently discard whatever a future provider adds, and the failure is asymmetric — refusing costs a slower interrupt (today's behavior), while wrongly discarding destroys work the model already did and the host may already have rendered.

**Stricter than the seal gate about server-side tools.** A seal may treat a settled `web_search_tool_result` as harmless, since the answer stays on the message. A restart would re-run and re-bill that search, so any tool machinery refuses.

**The wake channel.** `StreamPreemption.subscribe` is a *hint*: it wakes the SDK, which re-reads `shouldPreempt` as the sole authority, so the level-triggered contract is unchanged and spurious wakes are free. It is also read once per attempt — before the provider request is issued — because a request armed during setup or on a fallback-replaced attempt is already true and will never produce another wake. A self-rescheduling `unref`'d timer covers the remainder of the grace, since a silent provider yields no chunk to poll on.

**Lane keying.** A discard returns no message, so the model node cannot read `response_metadata.preempted` to learn a boundary is owed. The carrier is keyed by agent for the same reason `pendingPreemptReturn` is — one graph instance serves every parallel agent, and a single flag would let whichever lane finished first consume another lane's boundary. Threaded through `tryFallbackProviders` too, so a fallback-served attempt records into the same lane.

**Empty boundary after a restart re-issues the call.** A seal cannot self-loop there (a trailing model turn with no new input is dropped by some Gemini models and read as prefill by Anthropic), but a discard left graph state exactly as the node found it, so returning to the node simply re-issues the original call. A cancelled interrupt must not cost the user their answer. Bounded by the same seal budget.

**Reasoning predicate deduplicated.** `isReasoningContentBlock` moves to `src/messages/core.ts` and is shared with `src/stream.ts`. The stream classifier and the preempt gate must agree on what counts as reasoning, or a turn the host renders as thinking-only could look like a visible answer to the gate.

## Host contract

Probe `HOOK_PREEMPT_RESTART_CAPABLE` — deliberately separate from `HOOK_PREEMPT_BOUNDARY_CAPABLE`. An SDK with only the boundary still accepts the request and still labels the chip "interrupting"; it simply cannot act until the model starts writing, which is the difference users actually feel.

A host rendering from the stream should watch for `preemptDiscarded` on the synthetic `CHAT_MODEL_END`: a seal's content survives into the next prompt, a discard's does not.

## Test plan

- `src/specs/preemptRestart.test.ts` — four end-to-end scenarios through a real `Run`: silent provider discarded and re-issued with the steer appended (asserting no assistant message between the two user turns), thinking-only turn past the window, seal preferred when text arrives inside it, and empty boundary re-issuing the call.
- `src/llm/preempt.test.ts` — 48 cases over the gate and the grace policy.
- `preemptSeal.test.ts`, `Graph.preemptSignal.test.ts`, `preemptBoundary.test.ts` unchanged and green.
- Full suite: 233 suites / 4646 tests green (live-provider `llm.spec.ts` suites excluded — they fail identically on `main` without credentials).

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -19,6 +19,21 @@ export const DEFAULT_TOOL_TOKEN_MULTIPLIER = 1.4;
  * preemption-enabled run reserves.
  */
 export const DEFAULT_MAX_SEALS = 8;
+/**
+ * How long a preempt request waits for the turn to become sealable before it
+ * is allowed to discard the turn instead.
+ *
+ * Exists to keep a restart from stealing a seal that was about to happen: a
+ * model streaming reasoning is usually moments from its first text, and
+ * keeping that answer beats throwing the reasoning away. Only a request that
+ * outlives the window — a genuinely long thinking stretch — converts.
+ *
+ * Applies to a silent provider too. The accumulation a restart is judged
+ * against is what the consumer has seen, and the stream buffers a chunk ahead
+ * of it — so "nothing accumulated" can also mean "the first chunk is in
+ * flight", and converting on it would discard that chunk unseen.
+ */
+export const DEFAULT_PREEMPT_RESTART_GRACE_MS = 2_000;
 
 /**
  * Default ceiling on Stop-hook continuations within one Run. A blocking Stop
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -1463,6 +1463,21 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
    * another's turn.
    */
   pendingPreemptReturn = new Set<string>();
+  /**
+   * Agent IDs whose model turn a preempt DISCARDED rather than sealed. A
+   * discard returns no message, so the node cannot read
+   * `response_metadata.preempted` to learn a boundary is owed — this set is
+   * the only carrier.
+   *
+   * Keyed by agent for the same reason {@link pendingPreemptReturn} is: one
+   * graph instance serves every parallel agent, and a single flag would let
+   * whichever lane finished first consume another lane's boundary.
+   *
+   * Not derivable from `preemptSealInFlight`: an ordinary seal holds that slot
+   * too, and a claim that never reached its boundary would be indistinguishable
+   * from a discard.
+   */
+  private preemptRestartPending = new Set<string>();
 
   constructor(
     {
@@ -1712,6 +1727,7 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
   private resetPreemptTurnState(): void {
     this.preemptSealBudgetUsed = 0;
     this.preemptSealInFlight = false;
+    this.preemptRestartPending.clear();
     this.pendingPreemptReturn.clear();
   }
 
@@ -1808,6 +1824,20 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
   /** Releases the seal slot once its boundary has resolved, win or lose. */
   releasePreemptSeal(): void {
     this.preemptSealInFlight = false;
+  }
+
+  /** See {@link preemptRestartPending}. Called from the model attempt. */
+  notePreemptRestart(agentId: string): void {
+    this.preemptRestartPending.add(agentId);
+  }
+
+  /**
+   * Reads and clears the lane's mark in one step, so a boundary is dispatched
+   * exactly once per discard even though the model node runs again immediately
+   * after.
+   */
+  private consumePreemptRestart(agentId: string): boolean {
+    return this.preemptRestartPending.delete(agentId);
   }
 
   getPreemptStats(): t.PreemptStats {
@@ -3907,6 +3937,7 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
               {
                 request: preparedRequest,
                 context: this,
+                preemptAgentId: agentId,
               },
               invokeConfig
             )
@@ -4105,6 +4136,7 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
                 config: invokeConfig,
                 primaryError,
                 context: this,
+                preemptAgentId: agentId,
                 /**
                  * Lets the chain recognise a fallback overflow whose signature
                  * carries no reason of its own (Vertex AI's bare 400) and
@@ -4392,7 +4424,9 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
           { force: true }
         );
       }
+      const preemptRestarted = this.consumePreemptRestart(agentId);
       if (
+        preemptRestarted ||
         (responseMessage as AIMessageChunk | undefined)?.response_metadata
           .preempted === true
       ) {
@@ -4434,12 +4468,26 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
           return { messages: [...(result.messages ?? []), ...injected] };
         }
         /**
-         * Nothing to inject — the host cancelled or already drained. Do NOT
-         * self-loop: a trailing model turn with no new input is dropped by
-         * some Gemini models and read as prefill by Anthropic. Do NOT pretend
-         * the turn completed either; the answer really was cut short.
+         * Nothing to inject — the host cancelled or already drained.
+         *
+         * After a SEAL, do not self-loop: a trailing model turn with no new
+         * input is dropped by some Gemini models and read as prefill by
+         * Anthropic. Do not pretend the turn completed either; the answer
+         * really was cut short.
+         *
+         * After a RESTART there is no such trailing turn. The discard left
+         * graph state exactly as the model node found it, so returning to the
+         * node re-issues the same call — the only way the run can still
+         * produce an answer, and the honest outcome of an interrupt whose
+         * words were withdrawn before they landed. Bounded by the same seal
+         * budget, so a host stuck arming and cancelling cannot loop forever.
          */
         this.preemptEmptyBoundaries += 1;
+        if (preemptRestarted) {
+          this.pendingPreemptReturn.add(agentId);
+          this.cleanupSignalListener();
+          return result;
+        }
         this.preemptIncomplete = true;
       }
 
--- a/src/graphs/Graph.ts
+++ b/src/graphs/Graph.ts
@@ -2102,6 +2102,35 @@ export class StandardGraph extends Graph<t.BaseGraphState, t.GraphNode> {
    * unreachable — each step registers its calls before any completion can
    * reference it — and the mechanism was removed.)
    */
+  /**
+   * Closes the lane's open message step as `cancelled` when a preempt discards
+   * the turn that step belongs to.
+   *
+   * Without it the step is closed `completed` by the discard's own model-end,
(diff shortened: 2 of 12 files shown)


danny-avila:
@codex review

danny-avila:
Fixed in `9fbef0e`. Notes on each:

**P1 restart with no lane** — real, and my comment on that line claimed the opposite. The per-chunk trigger never consulted `restartLane`, so a seal-only host could resolve `restart`, return no message *and* record no lane, and end the turn empty with the steer still queued and a seal spent. The lane is now required for a restart, which doubles as the host's opt-in: a host that supplies no `subscribe` never signed up to have its turns discarded.

**P1 traces** — real, and it violates a stated `AGENTS.md` invariant. Keyed on the recorded model-run id rather than the error's shape, since every provider adapter raises its own cancellation error and matching on those would bind the invariant to whichever provider served the call. `handleLLMError` closes those as `controlFlow: 'PreemptRestart'`. Note the chunk-driven restart never aborts at all — it `break`s like a seal — so only the silent-provider path was affected.

**P2 skip invocation** — real. `evaluatePreemptRestart` reported whether *it* converted rather than whether a restart was decided, so a synchronous wake left the pre-call guard believing nothing happened. It now reports the decided state.

**P2 in-flight chunk** — real, and the grace window does not cover it: the window can be long past when the first text chunk arrives. The evaluator now declines while a chunk is mid-dispatch; the per-chunk poll runs immediately after the append with the full accumulation, so nothing is lost. Covered by construction rather than by a timing test — I could not arrange a deterministic hit on that await without a passing result proving nothing.

**P2 timer bound** — real. Clamped and chained, so an out-of-range grace is still reached, in several hops.

@codex review

danny-avila:
Fixed in `194f254`. All four were real.

**P1 stop consuming** — the abort is a request, not a guarantee, and the buffered-chunk case behaves the same from the loop's side. The loop now breaks the moment the decision is final.

Worth flagging how this got tested, because my first attempt was worthless: a model overriding `_streamResponseChunks` passed with *and* without the guard, since LangChain's base class enforces the signal above the generator. The committed test overrides `stream` so that handling never runs, leaving the SDK's loop as the only thing that can stop it — it yields 41 chunks and fails without the guard, stops early with it.

**P1 usage** — real, and it defeats the point of the round-one tracing fix: a generation closed with neither generations nor usage makes every restart free in cost accounting. Usage is now resolved at tear-down onto the same chunk the discard reports and rides in the marker, so the close carries a real generation. `synthesizeSealedUsage` short-circuits once usage exists, so the later discard path does not recompute it.

**P2 marker eviction** — right, and the fresh-evidence framing is the useful part: a count cap cannot tell a queued callback from a leak, and evicting a pending marker turns an expected restart back into a trace error. Age can tell them apart, so the bound is a TTL now, with a test that 500 concurrent records all survive and a stale one expires.

**P2 empty boundaries** — right. `emptyBoundaries` means a seal that ended the turn with nothing to resume from; a restart boundary that injects nothing reissues and produces a complete answer. No longer counted.

Full suite green: 233 suites / 4653 tests.

@codex review

danny-avila:
Fixed in `cebfbbaf`. All four real again.

**P1 per-chunk close** — the sharpest one. The two restart routes end the stream differently and I closed both the same way: only an abort makes LangChain close the run (through its error path, where the marker supplies usage), while breaking the iterator fires no callback at all. So every chunk-driven restart leaked an open Langfuse generation and lost its usage. The close now routes on `restartController.signal.aborted`, which is the exact record of which route ran — nothing else aborts it, and the run's signal is composed *into* the stream rather than into it.

**P2 seal counter** — right, and it contradicted two public contracts at once. Restarts take the same slot and the same budget (both end a turn, both cost a superstep) but now claim through `claimPreemptRestart` and count into a new `PreemptStats.restarts`. `seals` again means only turns that were kept.

**P2 expiry scheduling** — right, and the framing was the useful part: a host running preemption without tracing consumes *no* records, so the final burst is the common case rather than a corner one. A self-rescheduling `unref`'d timer now runs expiry independently and stops itself when the map empties.

**P2 natural end** — took this one. The grace guards against stealing a seal that was about to become possible, and a stream that ends while still discardable is proof none was coming. Leaving it to the terminal boundary replays a reasoning-only turn followed by the steer — the exact shape `canSealPreempt` refuses to build. Converting at the natural end costs no extra provider call since the stream is already over.

New coverage for the stats split, the non-truncating empty boundary, and the natural-end conversion. Full suite green: 233 suites / 4656 tests.

@codex review

danny-avila:
Fixed in `995df01`.

**P1 natural exhaustion** — correct, and it is my round-three fix biting back: that round added a *third* way for a restart to end the stream and I reused a two-way routing decision for it. The route is now explicit — `aborted` / `broke` / `exhausted` — and only `broke` closes natively, since it is the only one that fires no callback of its own. On the relabelling half I landed the opposite way from your suggestion: an exhausted run keeps its ordinary generation, because nothing was torn down. The generation really did complete; only the turn built on it is being discarded, and the close it already got is accurate.

**P2 control-flow output** — right, and it made trace interpretation depend on which trigger fired. `PREEMPT_RESTART_CONTROL_FLOW` moved to the preempt module and both routes emit it now.

**P2 grace clock** — right. Cleared at both observation points when the poll goes false, so a re-armed request inside one attempt gets its own window.

**P2 halted restarts** — right, same argument as the non-halting branch: nothing was preserved, so there is no truncated turn to report. `preemptIncomplete` still records that the answer never arrived.

Testing note: reaching the `broke` route deterministically needed a host whose wake never fires — with a live wake the abort route always wins the race, so the earlier version of that test was exercising a different path than it claimed. Both close-route tests verified to fail without the fix.

Full suite green: 233 suites / 4659 tests.

@codex review

danny-avila:
Fixed in `eec3596` — four taken, one partially.

**P1 duplicate usage** — right, and it is round four's fix finishing the job it started: stopping the duplicate *native* close left the synthetic one, which any host accumulating from `CHAT_MODEL_END` counts again. That route now sends only what the natural end could not say — the turn is discarded — with no usage attached.

**P1 concurrent failures** — right. The catch existed for one thing, this attempt's own abort, but keyed on `preemptAction` alone, so a breaker trip or a caller's cancellation in the same window was discarded and the node went on to consume the queued injection on a run that should have stopped.

**P2 late wake** — right; the stranded slot blocking every later preemption is the part that makes it more than cosmetic. The evaluator refuses once the attempt ends.

**P2 timer throw** — taken as containment rather than propagation. That frame has no invocation to reject into without threading a rejection handle through the attempt, and the cost of swallowing is one missed look by a predicate that is already broken.

**P2 marker TTL** — partially. Cleanup cannot be tied to the callback lifecycle because nothing here can observe when LangChain's queue drains, so I widened the window to five minutes and documented it as a leak bound rather than a deadline. If you have a lifecycle hook in mind that I have missed, say so and I will use it.

Testing note, since this bit twice more: both P1 tests were verified to fail without their fix, and my first two attempts did not. The usage one needs a `tokenCounter` on the run — without one the entire usage path silently no-ops — and has to watch the CUSTOM event channel, because the synthetic close never reaches `handleLLMEnd`. The cancellation one has to assert the boundary hook never ran; asserting only on the rejection passes either way.

Full suite green: 233 suites / 4661 tests. CI was 13/13 on the previous head.

@codex review

danny-avila:
Fixed in `765eb27`.

**P1 Responses tool output** — real, and the most serious finding of the cycle: not a trace or accounting problem but a repeated side effect. Reissuing after a completed `apply-patch` or shell call runs it again. Gated on both `tool_outputs` locations, with `response_metadata.output` inspected item by item — a reasoning-only Responses turn populates it too at natural completion, and that turn is exactly the one a restart should be able to discard.

**P1 ignored abort** — taken, but not as proposed, because the proposed fix is wrong. Switching the route to `broke` at the loop break assumes the error path did not run; the harness disproves that. Aborting can close the run *and* still hand over a chunk buffered before the abort landed, and the suggested change made the marker get dropped for a run whose `handleLLMError` may still be queued — trading an open generation for an error-labelled one. Since the run's state is not observable from here, both cut-short routes now attempt the native close and lean on `endSealedModelRun` degrading to the synthetic event when there is nothing to end. Correct either way; the expected-failure case is quiet so it does not warn on every interrupt, and the marker is kept.

**P2 blank text block** — right, and it is the string/array asymmetry: whitespace-only string content was already accepted, so a turn opening with `{type:'text',text:''}` was neither sealable nor discardable and would wait out the whole turn.

**P2 multi-handler markers** — right, and this codebase composes exactly that configuration in `endSealedModelRun`. Reads are non-consuming now; the sweep reclaims.

**P2 run steps** — took the marking half. `cancelOpenMessageStep` closes a cut-short attempt's open step as `cancelled` before the model-end can close it `completed`; an exhausted turn keeps `completed`, on the same reasoning as its generation. The test asserts only the reachable invariant — no step left `in_progress` — because a reasoning-only attempt does not open a MESSAGE_CREATION step in this harness, so I could not make the stronger assertion fail on purpose and will not claim coverage I do not have.

Full suite green: 233 suites / 4666 tests.

@codex review

PR #501, opened 2026-09-03
🫗 fix: Never Send Bedrock a Blank User Text Block

danny-avila:
## Problem

A user message whose text is empty is sent to Bedrock as a blank text block, and Bedrock
rejects the entire request:

```
ValidationException: The text field in the ContentBlock object at
messages.446.content.0 is blank. Add text to the text field, and try again.
```

Because every retry replays the same message, the conversation is wedged permanently. Seen
twice on a live instance in one day, both times from the same user action: attaching a
screenshot and sending without typing anything. The second occurrence happened with
LibreChat's own formatter guard (danny-avila/LibreChat#15530) already deployed — this sits
below that layer, so the guard could not reach it.

## Cause

`convertHumanMessageToConverseMessage` maps content without any emptiness check:

```ts
if (typeof msg.content === 'string') {
  userMessage.content = [{ text: msg.content }];          // '' -> [{ text: '' }]
} else if (Array.isArray(msg.content)) {
  userMessage.content = msg.content.map(...)               // { type:'text', text:'' } passes through 1:1
}
```

The **assistant** converter already handles both shapes — it guards `msg.content !== ''`
and, when filtering leaves no blocks, emits `BEDROCK_EMPTY_TEXT_PLACEHOLDER` with the comment
"Bedrock rejects an assistant message with no content blocks." The human path simply never
got the same treatment.

## Fix

Mirror the assistant path exactly:

- a blank string (`''` or whitespace-only) produces no text block
- blank text blocks are dropped from content arrays — an image-only send keeps its image
- if nothing remains, fall back to the existing `BEDROCK_EMPTY_TEXT_PLACEHOLDER`, since
  Bedrock rejects an empty user message just as it rejects a blank block

No new convention: the placeholder and the null guard are the ones the assistant converter
already uses.

## Tests

Four cases added to `message_inputs.test.ts`: empty string → placeholder, whitespace-only →
placeholder, image-only with a blank text block → image kept and no text block, and non-blank
text unchanged as a control. Verified against the original converter: the three guarding
cases fail and the control passes.

`npm run build` (tsdown + tsc) passes; `message_inputs.test.ts` + `llm.spec.ts` → **85 passed**;
ESLint clean.

### DIFF
--- a/src/llm/bedrock/llm.spec.ts
+++ b/src/llm/bedrock/llm.spec.ts
@@ -1349,7 +1349,9 @@ describe('convertToConverseMessages', () => {
       ]);
 
       const content = result.converseMessages[0].content!;
-      expect(content).toHaveLength(1);
+      expect(content).toHaveLength(2);
+      // Converse requires a text block alongside document blocks; it is appended last.
+      expect(content[1]).toEqual({ text: '_' });
       const documentBlock = expectDocumentBlock(content[0]);
       expect(documentBlock.document.format).toBe('pdf');
       expect(documentBlock.document.name).toBeDefined();
@@ -1397,7 +1399,9 @@ describe('convertToConverseMessages', () => {
       ]);
 
       const content = result.converseMessages[0].content!;
-      expect(content).toHaveLength(2);
+      expect(content).toHaveLength(3);
+      // Converse requires a text block alongside document blocks; it is appended last.
+      expect(content[2]).toEqual({ text: '_' });
       const firstDocument = expectDocumentBlock(content[0]);
       const secondDocument = expectDocumentBlock(content[1]);
       expect(firstDocument.document.name).toBeDefined();
--- a/src/llm/bedrock/utils/message_inputs.test.ts
+++ b/src/llm/bedrock/utils/message_inputs.test.ts
@@ -704,6 +704,26 @@ describe('convertToConverseMessages — blank user content', () => {
     });
   });
 
+  it('does not inject the placeholder into a merged user group', () => {
+    const { converseMessages } = convertToConverseMessages([
+      new HumanMessage('hello'),
+      new HumanMessage(''),
+    ]);
+    expect(converseMessages).toEqual([
+      { role: 'user', content: [{ text: 'hello' }] },
+    ]);
+  });
+
+  it('applies the placeholder when an entire user group is empty', () => {
+    const { converseMessages } = convertToConverseMessages([
+      new HumanMessage(''),
+      new HumanMessage('   '),
+    ]);
+    expect(converseMessages).toEqual([
+      { role: 'user', content: [{ text: '_' }] },
+    ]);
+  });
+
   it('keeps non-blank user text unchanged', () => {
     const { converseMessages } = convertToConverseMessages([
       new HumanMessage('hello'),
--- a/src/llm/bedrock/utils/message_inputs.test.ts
+++ b/src/llm/bedrock/utils/message_inputs.test.ts
@@ -688,6 +688,22 @@ describe('convertToConverseMessages — blank user content', () => {
     expect(content.some((block) => 'text' in block)).toBe(false);
   });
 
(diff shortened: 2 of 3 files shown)


danny-avila:
@codex review

danny-avila:
Both right. Fixed in bc8ce14e.

**One pass (P1).** The conversion is now a single `for...of` that routes text through the existing `appendSerializableBedrockTextBlock` and converts every other block inline — no intermediate array, no second traversal.

**Whitespace (P2).** Using that serializer also fixes this: a whitespace-only block is folded into the preceding text block rather than discarded, so `['hello', ' ', 'world']` serializes as `[{ text: 'hello ' }, { text: 'world' }]`. A blank block with nothing to fold into is dropped, and the placeholder still applies only when no serializable content remains. Test added for the folding case; the earlier four still pass.

`message_inputs.test.ts` + `llm.spec.ts` green; `npm run build` (tsdown + tsc) passes; ESLint clean.

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Both right. Fixed in 27c3c009.

**Narration (P1).** Removed; the loop and serializer call stand on their own.

**Placeholder vs. merging (P2).** Correct — the reducer concatenates consecutive `role: 'user'` messages, so a per-message placeholder leaked a synthetic `_` into a tool result or prompt it was merged with. The human converter now returns possibly-empty content, and the placeholder is applied by `finalizeUserMessage` only when a user group is complete: on the transition to a non-user message inside the same reduce pass, and once at the tail. A lone empty message still gets the placeholder; `['hello', '']` merges to `hello` alone; a group that is empty end-to-end gets exactly one placeholder. Two tests added for those cases; the earlier five still pass.

`message_inputs.test.ts` + `llm.spec.ts` green; `npm run build` (tsdown + tsc) passes; ESLint zero problems.

danny-avila:
@codex review

danny-avila:
Right: a promptless document upload arrives as blank text plus the document; once the blank is dropped, the document block survives, so the empty-group placeholder never fires, and Converse rejects a document without accompanying text. Fixed in df80a84e.

`finalizeUserMessage` now enforces both rules in one pass over the finalized group: an empty group gets the placeholder, and a group carrying a document block but no text block gets the placeholder appended. Still applied at group finalization, so a merged group that already has text from a neighbor is left alone. Test added: `[blank text, PDF file block]` serializes to `[document, { text: '_' }]`.

Two existing `llm.spec.ts` expectations changed: the document-conversion tests asserted document-only user messages (`toHaveLength(1)`/`(2)`), which is precisely the payload this finding says Converse rejects. Their filename-imputation assertions are untouched; they now also assert the trailing text block.

`message_inputs.test.ts` + `llm.spec.ts` green; `npm run build` (tsdown + tsc) passes; ESLint zero problems. The push gate fails closed on any test failure.

danny-avila:
@codex review

danny-avila:
Both right. Fixed in b33c3999.

**Finalization rescan (P1).** The human converter now records whether its single pass produced a top-level text block or a document block, keyed to the message in a `WeakMap`; merging ORs the two shapes; finalization reads the carried shape and is O(1). No content array is traversed after conversion.

**Leading whitespace (P2).** Whitespace with nothing before it to fold into is buffered until the next text fragment and prepended to it, so `[' ', 'hello']` serializes as `' hello'` and `[image, ' ', 'hello']` keeps its space. Whitespace after a text block still folds into it as before; whitespace left over at the end of a message has nothing to attach to and is dropped, as a blank block must be.

Tests added for the leading-whitespace and media-then-whitespace cases, plus a merged group where only a neighbor had text (no extra placeholder alongside the document). `message_inputs.test.ts` + `llm.spec.ts` green; `npm run build` passes; ESLint zero problems; the push gate fails closed on any test failure.

danny-avila:
@codex review

PR #509, opened 2026-09-04
🔦 fix: Log the Cause Behind the Sanitized Code API Authorization Error

danny-avila:
## The problem

`Code execution is not authorized. Verify access before trying again.` is produced by two entirely unrelated situations, and until now neither left a trace anywhere:

```ts
// the dynamic auth-header callback threw; the HTTP request is never sent
if (typeof authHeaders === 'function') {
  try { return await authHeaders(); }
  catch { throw new CodeApiRequestError(CODE_API_AUTHORIZATION_ERROR_MESSAGE); }
}

// the Code API answered with a real HTTP status
if (response.status === 401 || response.status === 403) {
  return CODE_API_AUTHORIZATION_ERROR_MESSAGE;
}
```

The bare `catch {}` discards the caught error entirely — not logged, not counted, not attached to the thrown error. Because it has no binding, this repo's own `preserve-caught-error` rule had nothing to bind to and stayed quiet, which is part of why it survived. The 401/403 branch likewise recorded neither the status it saw nor the body; `buildCodeApiHttpErrorMessage`'s `_method` and `_endpoint` parameters were unused, underscore-prefixed to say so. Downstream, `src/stream.ts` renders whatever survives as `Error: ${result.errorMessage}\n Please fix your mistakes.`, so the operator and the model see the same detail-free sentence.

## Why it matters

A production LibreChat instance hit this four times in a 1h45m window, in a tight cluster — both executions following each `create_file` call, zero failures anywhere else — and **it could not be attributed to either branch.** The upstream Code API logged nothing at any of the four timestamps. That is exactly what the callback path predicts, since in that branch no request is sent (this repo's own test asserts `expect(fetchMock).not.toHaveBeenCalled()`), and it is equally what a service that simply does not log rejected requests looks like. There is no way to tell from production data, so a recurring, user-visible failure has no root cause available without adding instrumentation and waiting for it to recur.

The failure is also actively misdirecting: the message points at credentials, so the agent's own recovery attempts — retrying, simplifying the script, switching tools — were all aimed at the wrong thing, and it lost code execution for the remainder of the turn immediately after producing a file it wanted to validate.

## What this changes

**Not the model-facing strings.** Every sentence the model can see is byte-identical, and the test asserting that neither the secret name nor the namespace reaches the model still passes. The sanitization is correct and stays. What changes is that the operator now gets what the model must not:

- `resolveCodeApiAuthHeaders` logs the callback failure with the original error, and re-throws `CodeApiRequestError` carrying it as `cause`. A host that wants to route it through its own logger now can.
- `buildCodeApiHttpErrorMessage` logs method, route, status, and the response body truncated to 512 chars. Its two parameters lose their underscores because they are finally used.
- 429 is the one routine rejection on this path, so it logs at `warn`; everything else at `error`.

`console.error` / `console.warn` follows what `CodeExecutor.ts` already does for its no-injected-files notice — these are free functions with six call sites across three modules and no logger in scope, and threading one through would be a far larger change for a diagnostic line.

## Tests

Four added under `the cause behind the sanitized authorization error`, all failing on `main`:

- the auth-header failure is logged and preserved as `cause`, while the message stays the sanitized one;
- an HTTP rejection logs its status, route and body while still returning the sanitized string;
- a 429 logs at warn, not error;
- a 4096-char error body is truncated rather than logged whole.

`npx jest src/tools/__tests__/` — 1115 passed, 23 skipped. `tsc -p tsconfig.json --noEmit` clean, `eslint` clean (the 3 warnings on `CodeExecutor.ts` are pre-existing on `main`), `check:circular-deps` clean. A few unrelated reflows in the diff are the repo's own pre-commit `prettier --write` acting on the touched files.

## Follow-up, not fixed here

The clustered failure that prompted this is still unattributed by design — that is the point of the change. Two adjacent LibreChat-side defects found in the same investigation are fixed in danny-avila/LibreChat#15593. Once this ships, one recurrence carries enough signal to close the original report.

### DIFF
--- a/src/tools/CodeExecutor.ts
+++ b/src/tools/CodeExecutor.ts
@@ -119,8 +119,7 @@ export const CodeExecutionToolSchema = {
   required: ['lang', 'code'],
 } as const;
 
-export const CODE_API_EXPECTED_PROFILE_HEADER =
-  'X-CodeAPI-Expected-Profile';
+export const CODE_API_EXPECTED_PROFILE_HEADER = 'X-CodeAPI-Expected-Profile';
 
 type SupportedLanguage = (typeof SUPPORTED_LANGUAGES)[number];
 
@@ -149,10 +148,28 @@ const SAFE_CODE_API_EXECUTION_ERROR_DETAILS: Readonly<
 };
 
 export class CodeApiRequestError extends Error {
-  constructor(message = CODE_API_UNAVAILABLE_ERROR_MESSAGE) {
+  constructor(
+    message = CODE_API_UNAVAILABLE_ERROR_MESSAGE,
+    options?: { cause?: unknown }
+  ) {
     super(message);
     this.name = 'CodeApiRequestError';
+    if (options?.cause !== undefined) {
+      this.cause = options.cause;
+    }
   }
+}
+
+const MAX_LOGGED_RESPONSE_BODY_CHARS = 512;
+
+/** Keeps a pathological error body from dominating the log line. */
+function truncateCodeApiResponseBody(responseBody: string): string {
+  if (responseBody === '') {
+    return '<empty body>';
+  }
+  return responseBody.length <= MAX_LOGGED_RESPONSE_BODY_CHARS
+    ? responseBody
+    : `${responseBody.slice(0, MAX_LOGGED_RESPONSE_BODY_CHARS)}... [truncated]`;
 }
 
 function getRetryAfterSeconds(responseBody: string): number | undefined {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Duplicate of #508, which landed first and is the better version — it additionally scrubs the logged response body through `redactSecrets`, which this PR did not. Folded the one idea worth keeping (`cause` on the re-thrown `CodeApiRequestError`) into a comment there.

PR #510, opened 2026-09-04
🎈 fix: Pass Through Search Chunks Without a Reranker

SSIG-IT:
`rerankerType: 'none'` is a schema-valid option, but it made web search answer with links and no text at all.

The chain: `createReranker` returns `undefined` for 'none' by design, so `getHighlights` bailed out on `!reranker` and returned `undefined`, so the source reached `expandHighlights` with content but no highlights -- and that function strips raw content it cannot expand. Every scraped page was discarded on the way to the model.

`getHighlights` now separates "nothing to rank" from "no reranker configured". Without a reranker the chunks pass through in their original order via `getDefaultRanking` -- the very ranking each reranker already falls back to when it fails, which is why `tool.ts` already logs "Using default ranking" for this case.

`getDefaultRanking` moves out of `BaseReranker` into an exported function so the pipeline can use it without instantiating a reranker; the protected method stays and delegates, so no subclass changes.

`highlights.ts` is deliberately untouched. "Raw scraped content must never leave this function" still holds for every other case: a source that truly has no highlights is still stripped, and even on the pass-through path the `content` field itself is still removed -- only highlights travel onward.

Note the size of what passes through: `getDefaultRanking` keeps the first `topK` candidates, so a source contributes roughly `topResults * chunkSize` characters (5 x 150 by default), not the whole page. That matches the existing fallback behaviour and keeps the chunk budget that stops a search from flooding the context. A test pins this down so the limit is visible rather than surprising.

Tests: 8 new in src/tools/search/no-reranker.test.ts; the full src/tools/search suite passes (250/250).

### DIFF
--- a/src/tools/search/no-reranker.test.ts
+++ b/src/tools/search/no-reranker.test.ts
@@ -1,0 +1,193 @@
+/**
+ * `rerankerType: 'none'` is a schema-valid opt-out, but it used to make the
+ * web search useless: `createReranker` returns `undefined` for it, so
+ * `getHighlights` returned `undefined`, so `expandHighlights` found a source
+ * with content but no highlights and stripped the content. The model received
+ * links and no text.
+ *
+ * The scraped chunks now pass through unscored via `getDefaultRanking` — the
+ * same ranking every reranker already falls back to when it fails.
+ */
+import type * as t from './types';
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #512, opened 2026-09-05
🧲 fix: Match MCP Server Filters Leniently in Tool Search

danny-avila:
## Summary

`tool_search`'s `mcp_server` filter compared the requested name against the tool's suffix with `===`, so a near-miss filtered everything out and reported it as an empty registry.

**The failure that prompted this.** An agent with 18 ClickHouse tools attached searched with:

```jsonc
tool_search { "query": "ClickHouse run select query", "mcp_server": "mcp-clickhouse" }
```

The server is keyed `clickhouse`, so its tools are named `run_select_query_mcp_clickhouse`; `mcp-clickhouse` is the *published package* name, which is what the model had read. Nothing matched, and the tool answered:

> `No tools available to search from MCP server(s): mcp-clickhouse. The tool registry is empty or no matching deferred tools are registered.`

The agent concluded the service was unavailable and hedged the answer it had been asked for, closing with "The ClickHouse code graph service was not available in this session." Its 18 ClickHouse tools were in the registry the whole time.

Two things were wrong: the filter had no tolerance for a name that means the same server, and the message conflated "registry is empty" with "your filter matched nothing" while never naming the servers that do exist.

## Changes

- **`resolveServerFilters`** maps requested names onto servers that actually have tools registered: exact match first, then case- and separator-insensitive, then with a leading `mcp` token dropped.
  - Separator leniency matters because a host rewrites characters it cannot put in a tool name — LibreChat turns `ClickHouse Cloud` into `ClickHouse_Cloud`, so a model asking for `clickhouse-cloud` is naming the same server.
  - Trying the three in that order keeps a deployment running both `github` and `mcp-github` unambiguous: each resolves to itself, never to the other.
- **The empty-result message now distinguishes three cases** it used to collapse into one: nothing is registered, the requested name matched nothing (and here are the servers that exist), or the named server genuinely has no tools. `available_mcp_servers` is added to the result metadata.
- Server listings (`mcp_server` with an empty query) format against the resolved names, so a lenient match lists the right server.
- `resolveServerFilters` and `canonicalizeServerName` are exported alongside the existing `isFromAnyMcpServer`, which is unchanged.

## Testing

`src/tools/__tests__/ToolSearch.test.ts` — **108 passed**.

Unit coverage for the resolver: exact names, the package-name case, case/separator differences, the `github` vs `mcp-github` disambiguation, unresolvable names, deduplication of filters that resolve to the same server, and mixed resolvable/unresolvable input.

Flow-level coverage driving `createToolSearch({ mode: 'local' })` against a small registry, which is what actually reproduces the bug:

- `mcp_server: 'mcp-clickhouse'` now returns the `clickhouse` tools and excludes the GitHub one.
- An unknown server name reports the available servers and no longer claims the registry is empty.

Those two polyfill `globalThis.crypto`, which Node 18 does not expose and LangChain's callback manager needs to mint a uuid before the tool body runs.

`npx tsc --noEmit` clean. ESLint reports 0 errors on the touched files (1 warning, also present on `main`).

## Note

Committed without a GPG signature at the author's request — the signing key could not be unlocked in the environment this was prepared in.

### DIFF
--- a/src/tools/ToolSearch.ts
+++ b/src/tools/ToolSearch.ts
@@ -189,14 +189,16 @@ function isFromMcpServer(toolName: string, serverName: string): boolean {
  * equal `☀️`. Collapsing separators is the only equivalence intended here, so
  * it is the only one performed.
  *
- * The uppercase/lowercase round trip approximates Unicode case folding, which
- * `toLowerCase` alone does not do: `ΟΣ` lowercases to `ος` with a final sigma
- * and would otherwise never match a registered `οσ`.
+ * Case is folded with `toLowerCase` alone. A round trip through uppercase
+ * would additionally equate a Greek final sigma with its medial form, but it
+ * also equates dotless `ı` with ASCII `i`, which Unicode case folding keeps
+ * apart — and a false match hands over another server's tools while a missed
+ * one only produces the message naming the servers that do exist. The safe
+ * direction is the one taken here.
  */
 function canonicalizeServerName(serverName: string): string {
   return serverName
     .trim()
-    .toUpperCase()
     .toLowerCase()
     .normalize('NFC')
     .replace(/[-_. ]+/g, '-');
@@ -1131,7 +1133,6 @@ ${mcpNote}${toolsListSection}
         let server: string | undefined;
         if (hasServerFilter) {
           server = extractMcpServerName(lcTool.name);
-          /** `tool_mcp_` yields an empty suffix that no filter can name. */
           if (server !== undefined && server !== '') {
             registeredServers?.add(server);
           } else {
--- a/src/tools/ToolSearch.ts
+++ b/src/tools/ToolSearch.ts
@@ -177,23 +177,29 @@ function isFromMcpServer(toolName: string, serverName: string): boolean {
 }
 
 /**
- * Reduces a server name to the form used for lenient matching: lowercase, with
- * separators and punctuation removed. A host rewrites characters it cannot put
- * in a tool name (LibreChat turns `ClickHouse Cloud` into `ClickHouse_Cloud`),
- * so a model asking for `clickhouse-cloud` is naming the same server.
+ * Canonical form used for lenient matching. It folds case and treats the
+ * separators a host may rewrite as equivalent — LibreChat turns
+ * `ClickHouse Cloud` into `ClickHouse_Cloud`, so a model asking for
+ * `clickhouse-cloud` is naming the same server.
  *
- * Letters, digits and combining marks are kept whatever their script, and the
- * result is composed so the same name written two ways agrees with itself.
- * Discarding any of them folds distinct names together: dropping non-ASCII
- * makes `éfoo` reduce to `foo`, and dropping marks makes `İfoo` — whose
- * lowercase is `i` plus a combining dot — reduce to `ifoo`. In both cases a
- * request for the plain name would return the other server's tools.
+ * **No character is ever discarded.** Every incorrect match this resolver has
+ * produced came from deleting characters, because deleting merges two distinct
+ * names onto one key: dropping non-ASCII made `éfoo` equal `foo`, dropping
+ * combining marks made `İfoo` equal `ifoo`, and dropping symbols made `❤️`
+ * equal `☀️`. Collapsing separators is the only equivalence intended here, so
+ * it is the only one performed.
+ *
+ * The uppercase/lowercase round trip approximates Unicode case folding, which
+ * `toLowerCase` alone does not do: `ΟΣ` lowercases to `ος` with a final sigma
+ * and would otherwise never match a registered `οσ`.
  */
 function canonicalizeServerName(serverName: string): string {
   return serverName
+    .trim()
+    .toUpperCase()
     .toLowerCase()
     .normalize('NFC')
-    .replace(/[^\p{L}\p{N}\p{M}]+/gu, '');
+    .replace(/[-_. ]+/g, '-');
 }
 
 /**
--- a/src/tools/ToolSearch.ts
+++ b/src/tools/ToolSearch.ts
@@ -187,15 +187,22 @@ function canonicalizeServerName(serverName: string): string {
 }
 
 /**
- * Drops a leading `mcp` token from an already-canonical name. Servers are
- * commonly keyed by the bare product (`clickhouse`) while the package that
- * implements them is called `mcp-clickhouse`, and a model that has read the
- * package name asks for that instead.
+ * Canonical form of a requested name with a leading `mcp` token removed, or
+ * undefined when it had none. Servers are commonly keyed by the bare product
+ * (`clickhouse`) while the package implementing them is `mcp-clickhouse`, and a
+ * model that read the package name asks for that instead.
+ *
+ * The boundary is checked on the original string, before canonicalization drops
+ * separators: `mcp-clickhouse` names the `clickhouse` server, but `mcparty` is
+ * simply a server called `mcparty` and must never resolve to `arty`.
  */
-function withoutMcpPrefix(canonical: string): string {
-  return canonical.startsWith('mcp') && canonical.length > 3
-    ? canonical.slice(3)
-    : canonical;
+function canonicalizeWithoutMcpPrefix(requested: string): string | undefined {
+  const match = /^mcp[-_. ]+(.+)$/i.exec(requested.trim());
+  if (match === null) {
+    return undefined;
+  }
+  const key = canonicalizeServerName(match[1]);
+  return key === '' ? undefined : key;
 }
 
 /**
@@ -254,7 +261,9 @@ function resolveServerFilters(
       resolved.push(...direct);
       continue;
     }
-    const stripped = canonical.get(withoutMcpPrefix(key));
+    const strippedKey = canonicalizeWithoutMcpPrefix(want);
+    const stripped =
+      strippedKey === undefined ? undefined : canonical.get(strippedKey);
     if (stripped !== undefined) {
       resolved.push(...stripped);
       continue;
@@ -909,12 +918,15 @@ function getDeferredToolsListing(
  * @param tools - Array of tool metadata from the server(s)
  * @param serverNames - The MCP server name(s)
  * @param nameFormat - Whether to show 'full' names (tool_mcp_server) or 'base' names (tool only)
+ * @param notes - Diagnostics about requested servers that were not searched;
+ * carried inside the JSON so the output stays parseable
  * @returns JSON string showing all tools grouped by server
  */
 function formatServerListing(
   tools: t.ToolMetadata[],
   serverNames: string | string[],
-  nameFormat: t.McpNameFormat = 'full'
+  nameFormat: t.McpNameFormat = 'full',
+  notes: string[] = []
 ): string {
   const servers = Array.isArray(serverNames) ? serverNames : [serverNames];
   const useFullName = nameFormat === 'full';
@@ -926,6 +938,7 @@ function formatServerListing(
         servers,
         total_tools: 0,
         tools_by_server: {},
+        ...(notes.length > 0 ? { notes } : {}),
         hint: 'No tools found from the specified MCP server(s).',
       },
       null,
@@ -960,6 +973,7 @@ function formatServerListing(
     servers,
     total_tools: tools.length,
     tools_by_server: toolsByServer,
+    ...(notes.length > 0 ? { notes } : {}),
     hint: `To use a tool, search for it by name (e.g., query: "${exampleToolName}") to load it.`,
   };
 
@@ -1223,6 +1237,7 @@ ${mcpNote}${toolsListSection}
               ...(hasServerFilter
                 ? { available_mcp_servers: availableServers }
                 : {}),
+              ...filterMetadata,
             },
           },
         ];
@@ -1231,14 +1246,17 @@ ${mcpNote}${toolsListSection}
       const isServerListing = hasServerFilter && query === '';
 
       if (isServerListing) {
+        /** The listing's contract is parseable JSON, so the diagnostics go
+         * inside the payload rather than in front of it. */
         const formattedOutput = formatServerListing(
           deferredTools,
           activeServers,
-          mcpNameFormat
+          mcpNameFormat,
+          filterNotes
         );
 
         return [
-          `${filterNotice}${formattedOutput}`,
+          formattedOutput,
           {
             tool_references: [],
             metadata: {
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
All three addressed in 13dcc11c.

**P1 — scanning every tool with no server filter.** Correct, and it was the hot path. `availableServers` was built with a map, a filter and a Set over the whole searchable registry on every call, while only the server-filtered path ever read it. Collection is now a single loop that produces the searchable subset and the server indexes together, and those indexes are only allocated when a filter was actually supplied. An unfiltered search now does one pass and no server bookkeeping.

**P2 — ambiguous canonical matches.** Real, and I had not considered two registered names colliding on one canonical form. `foo-bar` and `foobar` both reduce to `foobar`, and the map kept whichever came first, so `FOOBAR` was routed arbitrarily and the other server's tools vanished. Every candidate is returned now. I chose that over reporting ambiguity because an exact name still wins outright — so this only widens a request that was already indistinguishable — and a superset is recoverable by reading the tool names, whereas an arbitrary pick is silent. Two tests pin both halves: the inexact request gets both servers, the exact one gets only itself.

**P3 — resolving against only the searchable subset.** Also correct, and it reproduces the exact misdirection this branch exists to remove: with `onlyDeferred: true`, a server whose tools are all already loaded was reported as a name that does not exist. Resolution now runs against every registered server, and the outcome is split three ways — matched with searchable tools, matched with nothing left to search, or genuinely unknown. The message says which:

> `Registered with no tools left to search: slack. Server(s) with searchable tools: github.`

A test covers a registry holding a non-deferred Slack tool and a deferred GitHub one and asserts Slack is not described as unmatched.

```
Tests:       111 passed, 111 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`). CI was green on `a8f08f8f` (13 pass / 0 fail).

@codex review

danny-avila:
Four of five applied in e242c441; the fifth I believe is incorrect, evidence below.

**P1 — the unfiltered hot path.** You are right, and my "single pass" was worse than what it replaced: `Array.from` copied the registry, the loop filled a staging array, and `filter` + `map` walked that again. The staging array now exists only when a server filter does — that is the only case where the set to filter by is unknown until the whole registry has been read. An unfiltered search builds its results during the one traversal it already makes, with no copy and no intermediate array.

**P2 — idle diagnosis ordered after the empty-searchable fallback.** Correct, and it defeated the point of the previous round. A registry holding only loaded tools makes both conditions true, and the generic message won. The server diagnosis now runs first. New test: a registry containing a single non-deferred Slack tool, filtered on `slack`, reports `Registered with no tools left to search: slack.` and no longer claims the registry is empty.

**P2 — `available_mcp_servers` missing from the exported type.** Fixed; added as an optional field on `ToolSearchArtifact['metadata']` in `src/types/tools.ts`, so a typed consumer can read what the runtime already returns.

**P3 — empty server suffix.** Fixed. A tool named `tool_mcp_` yields `''`, which no filter can name and which `getAvailableMcpServers` already excludes; it is now skipped before reaching either index, so an unmatched search cannot print a bare `Server(s) with searchable tools: .` or emit `['']`.

**P1 — import ordering: not applied.** The finding says the `node:crypto` import is shorter than the `@jest/globals` one and must precede it. It is the longer line:

```
len=53  import { describe, it, expect } from '@jest/globals';
len=57  import { webcrypto as nodeWebcrypto } from 'node:crypto';
```

Shortest-to-longest therefore puts `@jest/globals` first, which is the current order. `node scripts/sort-imports.ts src/tools/__tests__/ToolSearch.test.ts` also reports `All 1 files already sorted`, and that script runs over these files in `lint-staged` on every commit here. Happy to change it if the intended rule is something other than rendered line length, but on both the stated rule and the repository's own tooling the existing order looks right.

```
Tests:       112 passed, 112 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

@codex review

danny-avila:
Both new findings applied in 21054f1f. Two of the four comments on this round are carried-over threads — notes on those at the end.

**P2 — empty canonical keys.** Good catch, and worse than described: `canonicalizeServerName` strips everything outside `[a-z0-9]`, so `---` and any Unicode-only name both reduce to `''`, and the index happily stored one under that key. A request for `!!!` — equally empty once canonicalized — then resolved to it and returned an unrelated server's tools. Empty canonical keys are now skipped on both sides: such a server stays reachable by its exact name only, and such a request is reported as unmatched. Test asserts `resolveServerFilters(['!!!'], ['---'])` resolves nothing while `['---']` still matches itself exactly.

**P2 — partial resolution.** Also correct. The diagnostics only existed on the empty-result path, so `['github', 'slack']` with GitHub deferred and Slack already loaded returned GitHub's results while the artifact echoed both names — reading as though both were searched. The unresolved and idle halves are now stated on every successful path:

> `Note: Registered with no tools left to search: slack.`

and carried in `unmatched_mcp_servers` / `idle_mcp_servers`. Test covers exactly that request shape and asserts both the GitHub result and the Slack notice.

---

**Carried-over: the artifact metadata type.** This was applied last round in `e242c441` — `src/types/tools.ts` gained `available_mcp_servers?: string[]`, and this round adds `unmatched_mcp_servers` and `idle_mcp_servers` beside it. The thread appears to have been re-raised against the older commit.

**Carried-over: the import ordering.** Still respectfully disagreeing, with the same evidence. Measured:

```
len=53  import { describe, it, expect } from '@jest/globals';
len=57  import { webcrypto as nodeWebcrypto } from 'node:crypto';
```

`node:crypto` is the **longer** line, so shortest-to-longest puts `@jest/globals` first — the current order. `node scripts/sort-imports.ts src/tools/__tests__/ToolSearch.test.ts` reports `All 1 files already sorted`, and that script runs over these files via `lint-staged` on every commit in this repo, including the three on this branch. If the intended rule is something other than rendered line length I will happily reorder — but I would rather not make the file disagree with the repository's own sorter on the strength of a measurement that appears to be inverted.

```
Tests:       114 passed, 114 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

@codex review

danny-avila:
All three new findings applied in 0a09dc99. The other four comments on this round are threads already resolved or answered; notes at the end.

**P2 — `mcp` prefix stripped without a boundary.** Correct, and the same failure class as the empty canonical key from round three: an unknown name quietly returning some other server's tools. `mcparty` began with those three letters and resolved to `arty`. The boundary is now checked on the original string, before canonicalization removes separators, so `mcp-clickhouse` and `mcp_clickhouse` still resolve to `clickhouse` while `mcparty` resolves to nothing. Tests pin all three.

**P2 — the notice broke the listing's JSON contract.** Right, and I should have caught this: `formatServerListing` returns JSON and the existing listing tests parse it, but I prefixed plain text in front of it. The diagnostics now travel inside the payload as a `notes` array. New test asserts `JSON.parse` succeeds on a partly-resolved listing and finds the skipped server in `notes`.

**P2 — `filterMetadata` missing from the empty return.** Fixed. The text explained the situation while the artifact did not, which defeats the purpose of exporting those fields.

---

**Already applied:** the artifact metadata type (`e242c441`), and both empty canonical keys and partial-resolution reporting (`21054f1f`). Those threads appear to be re-listed against the commits where they were raised rather than reopened.

**Still declining:** the import ordering. `node:crypto` is 57 characters against `@jest/globals`'s 53, so shortest-to-longest puts Jest first, which is the current order, and `node scripts/sort-imports.ts` reports the file already sorted on every commit including the four on this branch. Glad to change it if the intended rule is not rendered line length.

```
Tests:       116 passed, 116 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

@codex review

danny-avila:
Both new findings applied in da50f5fd.

**P2 — Unicode names folding onto ASCII ones.** Correct, and the fourth variant of the same mistake on this branch: canonicalization was *losing* characters rather than only dropping separators, so `éfoo` reduced to `foo` and a request for an unregistered `foo` returned another server's tools. Letters and digits of any script are now preserved (`[^\p{L}\p{N}]+` with the `u` flag) and only separators and punctuation are removed. The empty-key guard from round three still covers a name that is punctuation alone. Tests: `resolveServerFilters(['foo'], ['éfoo'])` resolves nothing, while `ÉFOO` still matches `éfoo` case-insensitively.

**P2 — search responses still outside the JSON.** Right, and my listing fix stopped one line short. `formatSearchResults` also returns `JSON.stringify(...)`, and both the local and code-interpreter success paths prefixed the notice in front of it. The diagnostics now ride inside that payload as a `notes` array, exactly as the listing does. `filterNotice` survives on one path only — the plain-text `No tools matched the pattern "…"` return, which is prose and never parsed. New test parses a partly-resolved search response and reads `notes` from it.

One thing I noticed while fixing this and deliberately did not change: the code-interpreter path already prefixes `warningMessage` ("Note: The provided pattern was converted to a literal search for safety.") in front of the same JSON, so that output is unparseable today whenever a pattern is sanitized. It predates this branch, so I have left it alone rather than widen the diff — happy to fix it here or in a follow-up, whichever you prefer.

```
Tests:       118 passed, 118 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

@codex review

danny-avila:
Applied in 1c7909c8, and verified against the runtime before changing anything rather than reasoning about it:

```
"İfoo" vs "ifoo"
   current: "ifoo" "ifoo"   -> collide
   fixed:   "i̇foo" "ifoo"  -> distinct
```

`İ`.toLowerCase() is `i` followed by U+0307, and the mark was being stripped, so the two names canonicalized identically and a request for a plain `ifoo` resolved to the other server. Marks are preserved now.

The same check surfaced a second inconsistency in the opposite direction, which the fix also settles: a decomposed `e` + combining acute name and a precomposed `éfoo` canonicalized *differently*, so one server written two ways failed to match itself. The result is composed with NFC, so both forms agree. Punctuation-only names still reduce to the empty string, so the round-three guard keeps them out of fuzzy matching.

Tests added for both directions: `ifoo` does not resolve to `İfoo`, and a decomposed request does resolve to its precomposed registration.

```
Tests:       120 passed, 120 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

Standing back for a moment: five of the findings across these six rounds have been the same class — a canonicalization that discarded characters and thereby merged two distinct identities, turning "found nothing" into "returned the wrong server's tools." Non-ASCII, combining marks, the missing `mcp` boundary, punctuation-only names, and two registered names sharing a canonical form. That is worth stating plainly in case it argues for a narrower design: the alternative is to drop fuzzy matching entirely and keep only the exact match plus the explicit `mcp-` prefix rule, which would address the original report — a model naming the package instead of the server — with far less surface. Happy to cut it back to that if you would prefer the smaller contract.

@codex review

danny-avila:
Both findings are correct, and rather than add a sixth guard I changed the approach — b1d86783.

Six of the findings on this branch have been one defect wearing different clothes: `éfoo` equalling `foo` (non-ASCII stripped), `İfoo` equalling `ifoo` (combining mark stripped), `❤️` equalling `☀️` (symbol stripped), every punctuation-only name collapsing to `''`, two registered names sharing a canonical key, and `mcp` stripped without a boundary. Every one came from the same act: **deleting characters**, which does not normalize a name, it merges two names onto one key.

So canonicalization no longer deletes anything. The equivalence this leniency was asked for is narrow — a separator written one way when the server uses another — so that is now the only equivalence performed:

```ts
serverName.trim().toUpperCase().toLowerCase().normalize('NFC').replace(/[-_. ]+/g, '-')
```

Case folded, NFC composed, separator runs collapsed to a single `-`. Nothing discarded. The uppercase/lowercase round trip is what approximates the Unicode case folding `toLowerCase` alone does not do, which is precisely what lets `ΟΣ` meet `οσ` — your second finding, fixed by the same change rather than a special case for sigma.

Verified pairwise against the runtime before committing:

| pair | before | now |
|---|---|---|
| `❤️` vs `☀️` | merged | distinct |
| `İfoo` vs `ifoo` | distinct | distinct |
| `éfoo` vs `foo` | distinct | distinct |
| `---` vs `!!!` | both `''` | distinct |
| `ΟΣ` vs `οσ` | **unmatched** | matched |
| `e`+acute vs `éfoo` | distinct | matched |
| `ClickHouse_Cloud` vs `clickhouse-cloud` | matched | matched |

The ambiguity handling from round two is retained and still earns its place — `foo-bar` and `foo_bar` genuinely share a key now — and its test was updated to a pair that actually collides under the new rule.

```
Tests:       125 passed, 125 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

@codex review

danny-avila:
All three applied in ba2a2f24.

**P2 — dotless `ı` merged with `i`.** Correct, and it was caused by the fix I made last round: the uppercase round trip I introduced to match a Greek final sigma also equates `ı` with `i`, which Unicode case folding keeps apart. I measured both candidates across the nine pairs this branch has accumulated:

| | false matches | missed matches |
|---|---|---|
| `toUpperCase().toLowerCase()` | **1** (`ıfoo` = `ifoo`) | 0 |
| `toLowerCase()` alone | **0** | 1 (`ΟΣ` vs `οσ`) |

Case is folded with `toLowerCase` alone now. The sigma capability is gone, and that is the right trade: a missed match produces the message naming the servers that do exist — the outcome this branch was written to provide — while a false match hands over another server's tools. Tests updated to assert the separation and to record the sigma limitation rather than pretend it works.

**P1 — narrating comment.** Removed; the `server !== undefined && server !== ''` check reads for itself.

**P1 — `unknown` in the crypto probe.** Given the concrete `Crypto` type instead.

```
Tests:       125 passed, 125 total
```

`tsc --noEmit` clean; eslint 0 errors on the touched files (1 warning, also on `main`).

---

Some perspective, since this is round eight. Seven of the findings have been one defect: a canonicalization rule that merged two distinct names onto one key. Non-ASCII, combining marks, symbols, punctuation-only names, the missing `mcp` boundary, shared canonical forms, and now a case-folding round trip. Each fix was correct and each revealed the next, which is the signature of a mechanism that is too general for its purpose.

What remains is deliberately small: exact match, then case and separator equivalence with nothing discarded, then an explicit `mcp-` prefix rule. That is the whole of what was asked for — a model naming `mcp-clickhouse` when the server is keyed `clickhouse`, or writing a hyphen where the host wrote an underscore.

If this round finds another character class that folds incorrectly, I would take that as the argument for removing the fuzzy layer altogether and keeping only the exact match plus the `mcp-` rule, which cannot merge anything. Worth deciding deliberately rather than discovering by attrition.

@codex review

PR #514, opened 2026-09-06
🫆 feat: Configure Langfuse Trace User Identity

danny-avila:
## Summary

Adds `userId` to `LangfuseConfig`. When a host sets it, every trace the run emits — the agent stream, titles, activity and reasoning labels, and phases — carries that identity as the Langfuse `userId` instead of `configurable.user_id`. Unset or blank keeps today's derivation, so existing hosts see no change.

Resolution lives in one place (`resolveLangfuseTraceUserId`), applied where handlers and propagated attributes are built, so every call site that already passes `langfuse` picks it up without touching the `configurable` contract that tools and MCP auth key on. Inherited identities (`inheritTraceIdentity`) stay untouched.

Host side: danny-avila/LibreChat#13529 — LibreChat resolves the user field an operator selected (email, username, IdP subject, …) and passes it here.

## Testing

```bash
npx jest src/specs/langfuse-config.test.ts   # 16 passed
npx tsc --noEmit
npx eslint src/langfuse.ts src/types/graph.ts src/specs/langfuse-config.test.ts
```

### DIFF
--- a/src/langfuse.ts
+++ b/src/langfuse.ts
@@ -895,6 +895,18 @@ function mergeLangfuseTags(
   return merged.length > 0 ? [...new Set(merged)] : undefined;
 }
 
+/**
+ * The trace's user identity: the host-configured `langfuse.userId` when
+ * present, else the caller's (normally `configurable.user_id`).
+ */
+export function resolveLangfuseTraceUserId(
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #547, opened 2026-09-16
🔄 fix: resume Anthropic context-window stops

upman:
## Summary

Resume once per agent/run after Anthropic's `model_context_window_exceeded` stop, preserving partial output and encrypted server-search results. Keep the raw streaming stop reason; pending client tools retain their normal route. A repeated stop sets the existing `output_truncated` halt reason.

```text
context stop → retain history + resume prompt → model
                                               ├─ answer → finish
                                               └─ context stop → halt
```

## Claude Code behavior

Local Claude Code 2.1.268 tests showed top-level context stops trigger a resume request with unchanged history, not compaction. Its nested WebSearch path instead returns links to the outer model. This follows the top-level continuation pattern, with a one-continuation safety bound and search history retained.

## Validation

| Check | Result |
| --- | --- |
| Regression tests | 292 passed, including streaming and tracing |
| Local tarball in LibreChat | Normal response: 1 call; context stop: 2 calls, persisted answer and aggregate usage; repeated stop: bounded at 2 |
| Gateway + Bedrock | Replayed a real context-stopped search response; LibreChat continued with encrypted history and received a coherent cited answer without another search |

The Bedrock continuation used a full-capacity model context: this verifies replay, not compaction of a still-full window. The tested LibreChat build does not yet map `output_truncated` to its persisted `unfinished` flag.


upman:
@codex review

upman:
@codex re-review

upman:
@codex re-review

PR #550, opened 2026-09-20
fix: cache tool loops with dynamic instructions

danny-avila:
## Summary

- add a rolling conversation-tail cache breakpoint when dynamic instructions are present, so growing tool loops stay cached
- reduce stable conversation-history markers from two to at most one, keeping the total budget at four breakpoints
- strip inherited Anthropic markers from the opening stable-prefix message so it is never used as an anchor
- preserve existing instruction placement, message ordering, the no-dynamic-instructions path, and Bedrock behavior

## Why

With prompt caching enabled and both stable and dynamic instructions present, `AgentContext` inserts the dynamic instructions near the latest human message and marks only the preceding history, skipping the normal conversation-tail marker. `Graph.ts` does not compensate, because it delegates message-marker placement to the system runnable whenever one exists.

For an isolated subagent, whose task is its only human message, that leaves the growing tool loop with no conversation breakpoint at all: every tool step re-reads the whole loop uncached. Primary-agent tool loops have the same gap after the latest user turn.

`addTailCacheControl()` strips existing markers from its input, so calling it on the assembled body would erase the stable-history breakpoint and could anchor an instruction-only carrier. Applying it only to the conversation suffix, before reassembly, keeps that boundary intact.

## Breakpoint budget

| Region | Maximum |
|---|---:|
| Tool definitions | 1 |
| Stable system instructions | 1 |
| Stable conversation history | 1 |
| Rolling conversation tail | 1 |
| **Total** | **4** |

Correct marker placement is necessary but not sufficient: reuse still depends on matching preceding content and on provider requirements.

Related: danny-avila/LibreChat#16044

## Tests

New `src/agents/__tests__/AgentContext.promptCache.test.ts` reproduced the missing tail marker before the fix (15 failed, 2 passed against unchanged production code). `src/agents/__tests__/AgentContext.test.ts` expectations were updated for one stable-history marker plus the rolling tail.

Coverage includes growing subagent loops and primary-agent history, Anthropic and OpenRouter, both cache TTL settings, changed dynamic instructions with preserved stable-prefix content, summaries and post-compaction tails, inherited and replayed markers, input immutability, disabled caching, the no-dynamic-instructions path, single-message ordering, synthetic and reasoning-only tail blocks, Anthropic tool-result marker serialization, OpenRouter tool-result marker preservation, and the four-breakpoint budget.

- focused cache and compatibility suites: 521 passed across 15 suites
- tracing suites: 203 passed across 11 suites
- `npx tsc --noEmit`
- focused ESLint and `npm run sort-imports:check` on the three changed files

Not run locally: package build, full suite, and live provider cache or billing measurements.


danny-avila:
@codex review

Please review the current PR head ce0cc1bebc0303ad0d00f8d2c451c5c57832cc9f. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

